PR #29, opened 2017-11-03
Adds CI/NuGet packaging

### LINKED ISSUE #8: Add Nuget Generation to build
@JunyiYi can we get nuget builds from the server? It will make it a lot easier to consume updates

Added this commit with a sample nuspec: https://github.com/Azure/azure-function-java-host/commit/c1b1a035bf2c296ae2098220e9bf9b3117ee9bd0

in build scripts:
`nuget pack Worker.nuspec -Properties version=10000` (insert a $version$ into the nuspec version string)

We can set up the build to automatically publish to myget as well.

Currently consuming here: https://github.com/Azure/azure-webjobs-sdk-script/blob/core/src/WebJobs.Script/WebJobs.Script.csproj#L18

Nuget generation for java is currently a manual step.

christopheranderson:
Adds:
 - nuspec file
- appveyor.yml

fixes #8

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,33 @@
+version: '{build}'
+
+image: Visual Studio 2017
+
+pull_requests:
+  do_not_increment_build_number: true
+
+branches:
+  only:
+    - dev
+    - master
+
+environment:
+  matrix:
+    - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
+
+install:
+  - cmd: echo %JAVA_HOME%
+  - ps: Get-Command mvn
+  - cmd: mvn -v
+  - ps: Get-Command nuget
+
+build_script:
+  - mvn clean install -DskipTests
+  - ps: nuget pack ./tools/AzureFunctionsJavaWorker.nuspec -Properties versionsuffix=$env:APPVEYOR_BUILD_NUMBER
+
+artifacts:
+  - path: '*.nupkg'
+  - path: 'azure-functions-java-worker/target/**.jar'
+  - path: 'azure-functions-java-core/target/**.jar'
+
+cache:
+  - C:\maven\
--- a/tools/AzureFunctionsJavaWorker.nuspec
+++ b/tools/AzureFunctionsJavaWorker.nuspec
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+  <metadata>
+    <id>Microsoft.Azure.Functions.JavaWorker</id>
+    <title>Microsoft Azure Functions Java Worker</title>
+    <tags>java azure-functions azure</tags>
+    <version>1.1.0-beta1-$versionsuffix$</version>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <projectUrl>https://github.com/Azure/azure-functions-java-worker</projectUrl>
+    <licenseUrl>https://github.com/Azure/azure-functions-java-worker/blob/master/LICENSE</licenseUrl>
+    <iconUrl>https://raw.githubusercontent.com/Azure/azure-functions-cli/master/src/Azure.Functions.Cli/npm/assets/azure-functions-logo-color-raster.png</iconUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Microsoft Azure Functions Java Worker</description>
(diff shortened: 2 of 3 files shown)


tools/AzureFunctionsJavaWorker.nuspec
```diff
@@ -0,0 +1,23 @@
...
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <projectUrl>https://github.com/Azure/azure-functions-java-worker</projectUrl>
+    <licenseUrl>https://github.com/Azure/azure-functions-java-worker/blob/master/LICENSE</licenseUrl>
+    <iconUrl>https://raw.githubusercontent.com/Azure/azure-functions-cli/master/src/Azure.Functions.Cli/npm/assets/azure-functions-logo-color-raster.png</iconUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Microsoft Azure Functions Java Worker</description>
+    <copyright>Copyright (c) Microsoft Corporation. All rights reserved.</copyright>
+    <contentFiles>
+      <files include="../azure-functions-java-worker/target/azure-functions-java-worker-1.0-SNAPSHOT.jar" buildAction="None" copyToOutput="true" flatten="false"/>
```
JunyiYi: So we need to update this (1.0-SNAPSHOT) for every new version (maybe to 1.0.0-beta-2 for the next release)?

tools/AzureFunctionsJavaWorker.nuspec
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+  <metadata>
+    <id>Microsoft.Azure.Functions.JavaWorker</id>
+    <title>Microsoft Azure Functions Java Worker</title>
+    <tags>java azure-functions azure</tags>
+    <version>1.1.0-beta1-$version$</version>
```
JunyiYi: What does this version mean? Is it required to match the Java worker JAR version?

christopheranderson: This is the version of the NuGet. It's not required to match the .jar. We can version them independently, if we wish.

tools/devdoc.md
```diff
@@ -0,0 +1,20 @@
+# Build and publishing
+
+In order for the Java Worker to get loaded into the runtime, today we use a NuGet to package the jars for deployment.
+
+## AppVeyor
+
+We're using App Veyor for CI. It builds the Java Core and Worker modules and then packages them into a NuGet for consumption by the Azure Functions Script Host.
+
+CI runs on every PR and nightly on `dev` and `master`
```
JunyiYi: Is it OK to assume that only the changes in `master` branch will be packed into the latest runtime.

tools/AzureFunctionsJavaWorker.nuspec
christopheranderson: Let me try something to make this more generic.

tools/devdoc.md
christopheranderson: I did dev and master since master is usually the "released" bits and dev and the "to be released" bits. It's what the rest of the Functions repos do.

JunyiYi: Sure. So does that mean we may do experiments in `dev` branch, and the user won't notice that until we pushed them to `master`?

christopheranderson: This just produces CI builds for us, I didn't automate publishing, so users won't see anything we do unless we publish.

appveyor.yml
```diff
@@ -0,0 +1,33 @@
...
+
+install:
+  - cmd: echo %JAVA_HOME%
+  - ps: Get-Command mvn
+  - cmd: mvn -v
+  - ps: Get-Command nuget
+
+build_script:
+  - mvn clean install -DskipTests
+  - ps: nuget pack ./tools/AzureFunctionsJavaWorker.nuspec -Properties version=$env:APPVEYOR_BUILD_NUMBER
```
fabiocav: Would recommend renaming the property to `versionsuffix` (that's what we use in the other projects) or something the better describes this, as the actual version is the concatenation of major.minor and this property

christopheranderson: Should I make version a modifiable field as well?

christopheranderson: I could just make the whole thing $version$ and assemble the string via a PS expression, right?

fabiocav: Up to you. Usually, the major minor go in the nuspec and we just take the suffix as a parameter.

Issue #58, opened 2018-02-21
Event Grid trigger support


christopheranderson:
Java should support Event Grid now that Event Grid is GA.

kevinhillinger:
This is now [in the source](https://github.com/Azure/azure-functions-java-worker/blob/dev/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/EventGridTrigger.java). I believe this issue can be closed.

JunyiYi:
Annotation is published, but the tooling has not been updated. @pragnagopa , you can close this as soon as the functions extension tooling is published.

paulbatum:
This issue should have been closed once the change was committed. I'll let @pragnagopa comment on whether this is actually released or not yet.

PR #59, opened 2018-02-22
add EventGridTrigger support

christopheranderson:
fixes https://github.com/Azure/azure-functions-java-worker/issues/58

(Still need to test this, so do not merge it yet :))

### DIFF
--- a/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/EventGridTrigger.java
+++ b/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/EventGridTrigger.java
@@ -1,26 +1,3 @@
-/*
-[JsonProperty(PropertyName = "topic")]
-        public string Topic { get; set; }
-
-        [JsonProperty(PropertyName = "subject")]
-        public string Subject { get; set; }
-
-        [JsonProperty(PropertyName = "data")]
-        public JObject Data { get; set; }
-
-        [JsonProperty(PropertyName = "eventType")]
-        public string EventType { get; set; }
-
-        [JsonProperty(PropertyName = "publishTime")]
-        public DateTime PublishTime { get; set; }
-
-        [JsonProperty(PropertyName = "eventTime")]
-        public DateTime EventTime { get; set; }
-
-        [JsonProperty(PropertyName = "id")]
-        public string Id { get; set; }
-*/
-
 /**
  * Copyright (c) Microsoft Corporation. All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for
--- a/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/EventGridTrigger.java
+++ b/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/EventGridTrigger.java
@@ -1,0 +1,42 @@
+/*
+[JsonProperty(PropertyName = "topic")]
+        public string Topic { get; set; }
+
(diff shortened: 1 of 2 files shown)


JunyiYi:
It would be better to also upgrade the following versions to `1.0.0-beta-3`:
* https://github.com/Azure/azure-functions-java-worker/blob/dev/azure-functions-java-core/pom.xml#L5
* https://github.com/Azure/azure-functions-java-worker/blob/dev/azure-functions-java-worker/pom.xml#L47

And update the version below to `1.0.0-beta-4` as well:
* https://github.com/Azure/azure-functions-java-worker/blob/dev/azure-functions-java-worker/pom.xml#L5

kevinhillinger:
Looking into supporting the binding in the maven plugin, I noticed that there seems to be an implicit contract being created where `name()` and `dataType()` is the minimum definition for a binding.

see [BaseBinding](https://github.com/Microsoft/azure-maven-plugins/blob/master/azure-functions-maven-plugin/src/main/java/com/microsoft/azure/maven/function/bindings/BaseBinding.java)

I noticed this PR has EventGridTrigger is missing them.

Since annotation type definitions can't implement explicitly, I'm proposing that at least we get a unit test in place to verify that `name()` and `dataType()' are part of any annotation under `functions` package, particularly since we're building up a lot of efferent coupling in Functions.

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/EventGridTrigger.java
```diff
@@ -0,0 +1,19 @@
...
+
+package com.microsoft.azure.serverless.functions.annotation;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER, ElementType.METHOD})
```
xscript: Triggers should be only applied to `ElementType.Parameter`.

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/EventGridTrigger.java
```diff
@@ -0,0 +1,19 @@
...
+public @interface EventGridTrigger {
+
```
xscript: As @kevinhillinger pointed out, `name` and `dataType` are required fields.

xscript:
@kevinhillinger You are right. Ideally, we should put common fields to a base class/annotation. But Java doesn't allow annotation inheritance. So we have to put them in every single annotation.

As to you proposal, we also have some non-binding annotations:
-  [FunctionName](https://github.com/Azure/azure-functions-java-worker/blob/dev/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/FunctionName.java)
- [StorageAccount](https://github.com/Azure/azure-functions-java-worker/blob/dev/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/StorageAccount.javal)
-  [BindingName](https://github.com/Azure/azure-functions-java-worker/blob/dev/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/BindingName.java)

You should put them into a white-list when adding unit tests for annotations.

kevinhillinger:
@christopheranderson can these changes be incorporated into the PR? There's situations coming up for being able to use Event Grid trigger, and this will take it across the finish line.

PR #72, opened 2018-03-20
A few improvements to the JavaDoc for Azure Functions / Java

JonathanGiles:
A first round of suggested JavaDocs. Please let me know what is correct and what is not, so that I may tweak it to better suit your needs. A lot of the text is guesswork by me and might need tightening up - please let me know where this is the case and I will do it (or edit the PR directly).

### DIFF
--- a/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/HttpRequestMessage.java
+++ b/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/HttpRequestMessage.java
@@ -22,8 +22,8 @@ public interface HttpRequestMessage<T> {
     URI getUri();
 
     /**
-     * Returns the name of the method called.
-     * @return the name of the method called.
+     * Returns the HTTP method name, such as "GET" and "POST".
+     * @return the HTTP method name, such as "GET" and "POST".
      */
     String getMethod();
 
--- a/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/HttpRequestMessage.java
+++ b/azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/HttpRequestMessage.java
@@ -3,12 +3,56 @@ package com.microsoft.azure.serverless.functions;
 import java.net.URI;
 import java.util.Map;
 
+/**
+ * An HttpRequestMessage instance is provided to Azure functions that use
+ * {@link com.microsoft.azure.serverless.functions.annotation.HttpTrigger HTTP Triggers}. For an example of how to use
+ * the http functionality of Azure Functions, refer to the example in the
+ * {@link com.microsoft.azure.serverless.functions.annotation.HttpTrigger}
+ *
+ * @see com.microsoft.azure.serverless.functions.annotation.HttpTrigger
+ * @see HttpResponseMessage
+ * @param <T> The type of the body content that is expected to be received as part of this HTTP request.
+ * @since 1.0.0
+ */
 public interface HttpRequestMessage<T> {
+    /**
+     * Returns the URI that was called that resulted in this HTTP request being submitted.
+     * @return the URI that was called that resulted in this HTTP request being submitted.
+     */
     URI getUri();
+
+    /**
+     * Returns the name of the method called.
+     * @return the name of the method called.
+     */
     String getMethod();
+
+    /**
+     * Returns a map of headers that were contained within this HTTP request.
+     * @return a map of headers that were contained within this HTTP request.
+     */
     Map<String, String> getHeaders();
+
+    /**
+     * Returns a map of query parameters that were included with this HTTP request.
+     * @return a map of query parameters that were included with this HTTP request.
+     */
     Map<String, String> getQueryParameters();
+
+    /**
+     * Returns any body content that was included with this HTTP request.
+     * @return any body content that was included with this HTTP request.
+     */
     T getBody();
 
+    /**
+     * Generates a {@link HttpResponseMessage} instance containing the given HTTP status code and response body.
+     * Additional headers may be added by calling appropriate methods on {@link HttpResponseMessage}.
+     *
+     * @param status The HTTP status code to return to the caller of the function.
+     * @param body The body content to return to the caller of the function.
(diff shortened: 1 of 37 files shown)


azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/HttpRequestMessage.java
```diff
@@ -3,12 +3,56 @@
...
+ */
 public interface HttpRequestMessage<T> {
+    /**
+     * Returns the URI that was called that resulted in this HTTP request being submitted.
+     * @return the URI that was called that resulted in this HTTP request being submitted.
+     */
     URI getUri();
+
+    /**
+     * Returns the name of the method called.
```
JunyiYi: Returns the HTTP method name, such as "GET" and "POST".

JonathanGiles:
Pull request updated with additional content. Still nowhere near complete, but it is nearing the point where I can't add much more as my experience is limited to only some of the triggers. Perhaps consider accepting and merging this at this point?

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/HttpTrigger.java
```diff
@@ -11,18 +11,108 @@
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * <p>The HttpTrigger annotation is applied to Azure functions that will be triggered by a call to the HTTP endpoint that
+ * the function is located at. The HttpTrigger annotation should be applied to a method parameter of type
+ * {@link com.microsoft.azure.serverless.functions.HttpRequestMessage}. For example:</p>
```
JunyiYi: The HttpTrigger annotation could be applied to a method parameter of various types, such as int, String, Optional&lt;String&gt;, any POJO types and HttpRequestMessage&lt;T&gt;.

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/HttpTrigger.java
```diff
@@ -11,18 +11,108 @@
...
+     * <p>Defines the route template, controlling which request URLs your function will respond to. The default value if
+     * no route is provided is the function name specified in the {@link FunctionName} annotation, applied to each
+     * Azure Function.</p>
+     *
+     * <p>By default when you create a function for an HTTP trigger, or WebHook, the function is addressable with a
+     * route of the form {@code http://<yourapp>.azurewebsites.net/api/<funcname>}. You can customize this route using
+     * this route property. For example, a route of {@code "products/{category:alpha}/{id:int}"} would mean that the
+     * function is now addressable with the following route instead of the original route:
+     * {@code http://<yourapp>.azurewebsites.net/api/products/electronics/357}, which allows the function code to support
+     * two parameters in the address: category and id.</p>
```
JunyiYi: You can use `@BindingName("category") String cat` and `@BindingName("id") int id` as your function method parameters to get those route values.

JonathanGiles: Thanks - adding a sample here to explain this was on my todo list for today, but first I had to write some code! I'll do that today and then update this text.

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/HttpTrigger.java
JonathanGiles: Thanks for clarifying. Is there any documentation that you know of that clarifies this in more detail - when should a developer use one type over another, etc?

JonathanGiles:
JavaDoc updated with additional content.

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/HttpTrigger.java
```diff
@@ -11,18 +11,142 @@
...
 import java.lang.annotation.Target;
 
+/**
+ * <p>The HttpTrigger annotation is applied to Azure functions that will be triggered by a call to the HTTP endpoint that
+ * the function is located at. The HttpTrigger annotation should be applied to a method parameter of one of the following
+ * types:</p>
+ *
+ * <ul>
+ *     <li>{@link com.microsoft.azure.serverless.functions.HttpRequestMessage HttpRequestMessage&lt;T&gt;}</li>
+ *     <li>int</li>
```
JunyiYi: Replace "int" and "String" with "any native Java types such as int, String, byte[]"

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/HttpTrigger.java
```diff
@@ -11,18 +11,142 @@
...
+ *     <li>String</li>
+ *     <li>Optional&lt;String&gt;</li>
```
JunyiYi: Replace `Optional<String>` with `nullable values using Optional<T>`

azure-functions-java-core/src/main/java/com/microsoft/azure/serverless/functions/annotation/HttpTrigger.java
```diff
@@ -11,18 +11,142 @@
...
+ *     <li>Any POJO type</li>
+ * </ul>
+ *
```
JunyiYi: These restrictions also apply to the return type of the method. The only difference is that you should use `HttpResponseMessage` instead of `HttpRequestMessage`.

JunyiYi:
Thanks @JonathanGiles for the valuable contributions. Let's just refine the existing javadoc in this PR now because there has been a lot of changed files which is very hard to review. We might create new PRs for other newly added javadoc.

JonathanGiles:
Small tweaks based on feedback have been pushed. Thanks for reviewing - can't wait to see this merged.

JonathanGiles:
Can this be merged?

PR #78, opened 2018-03-27
Update readme.md file to remove shade advice

JonathanGiles:
Shade advice is no longer recommended practice, so removing from the readme.md file before it leads to additional confusion.


README.md
```diff
@@ -33,46 +33,18 @@ public class MyClass {
 
 ### Including 3rd Party Libraries
```
JunyiYi: It would be better to mention: If you are generating your project using the latest Azure functions java archetype, the dependencies section is automatically created for you.

PR #79, opened 2018-04-04
Increase both core and worker version since both of them will be updated.

JunyiYi:
* Increase core version: annotations will be changed.
* Increase worker version: reference to the new core.


JunyiYi:
@pragnagopa , if you need to increase the versions in the future, you may just reference this pull request to find all versions defined in this repository.

Issue #93, opened 2018-04-26
GRPC message length should be taken in and set as input parameter


mhoeger:
Set send max send and receive message sizes on java client. Currently broken without fix because of following error:

`[SEVERE] {Application.parseCommandLine}: org.apache.commons.cli.UnrecognizedOptionException: Unrecognized option: --grpcMaxMessageLength`

pragnagopa:
Fixed in Commit # [e6c21f8c0d825fd7397d84e42fecd911a14061b1](https://github.com/Azure/azure-functions-java-worker/commit/e6c21f8c0d825fd7397d84e42fecd911a14061b1)

PR #94, opened 2018-04-26
HOTFIX: Add grpc max message length argument.

JunyiYi:
hotfix to unblock the worker publish.

### DIFF
--- a/azure-functions-java-worker/src/main/java/com/microsoft/azure/webjobs/script/JavaWorkerClient.java
+++ b/azure-functions-java-worker/src/main/java/com/microsoft/azure/webjobs/script/JavaWorkerClient.java
@@ -22,7 +22,11 @@ import com.microsoft.azure.webjobs.script.rpc.messages.*;
 public class JavaWorkerClient implements AutoCloseable {
     public JavaWorkerClient(IApplication app) {
         WorkerLogManager.initialize(this, app.logToConsole());
-        this.channel = ManagedChannelBuilder.forAddress(app.getHost(), app.getPort()).usePlaintext(true).build();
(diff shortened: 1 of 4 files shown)


JunyiYi:
This resolves #93 .

azure-functions-java-worker/src/main/java/com/microsoft/azure/webjobs/script/JavaWorkerClient.java
```diff
@@ -22,7 +22,11 @@
 public class JavaWorkerClient implements AutoCloseable {
     public JavaWorkerClient(IApplication app) {
         WorkerLogManager.initialize(this, app.logToConsole());
-        this.channel = ManagedChannelBuilder.forAddress(app.getHost(), app.getPort()).usePlaintext(true).build();
+        ManagedChannelBuilder<?> chanBuilder = ManagedChannelBuilder.forAddress(app.getHost(), app.getPort()).usePlaintext(true);
+        if (app.getMaxMessageSize() != null) {
+            chanBuilder.maxInboundMessageSize(app.getMaxMessageSize());
```
pragnagopa: For node and c# we need to set two values: max send  size, max receive size. Java probably needs maxOutboundMessageSize?

JunyiYi: AFAIK, Java GRPC has [unlimited default outbound message size](https://github.com/grpc/grpc-java/blob/9fac15d4f8fb09106fc4a80d761afbe1594b4d0c/core/src/main/java/io/grpc/internal/MessageFramer.java#L73).

PR #98, opened 2018-04-27
Integration tests

kevinhillinger:
initial set of integration tests for the worker that represents the team's initial set of smoke tests. The intent is to get a base of tests going, and add continue the work.

Currently, the pre-integration-test phase will start up the core tools host on the same machine and then tear it down once the tests are complete. I think this is a good start, but it assumes that all the scaffolding is in place.

Libraries that were pulled in:
* rest-assured
* assertJ

FYI, the process termination for windows (to stop the core tools) is not implemented in the POM yet, but I will be adding that as part of the PR.


JunyiYi:
Thanks @kevinhillinger for contributing this. But we will not introduce any integration tests in this repository, instead, we will do it in [Azure Functions Host](https://github.com/Azure/azure-functions-host/tree/dev/test) repository. We only do unit tests here which has already been introduced.

kevinhillinger:
@JunyiYi Thanks for clarifying. I will work to translate, and close this PR.

PR #104, opened 2018-05-15
Delete annotations from worker repo

pragnagopa:


pragnagopa:
@JunyiYi - Updated PR and tested it builds locally when I installed azure-functions-java-library in local maven repository

PR #106, opened 2018-05-22
Consume latest proto file

pragnagopa:
- Added subtree from https://github.com/azure/azure-functions-language-worker-protobuf


pragnagopa:
cc @mhoeger

src/main/java/com/microsoft/azure/functions/binding/RpcComplexDataSources.java
```diff
@@ -81,7 +81,8 @@
         super(name, null, HTTP_DATA_OPERATIONS);
         this.httpPayload = value;
         this.bodyDataSource = BindingDataStore.rpcSourceFromTypedData(null, this.httpPayload.getBody());
-        this.fields = Arrays.asList(this.httpPayload.getHeadersMap(), this.httpPayload.getQueryMap(), this.httpPayload.getParamsMap());
+        this.fields = Arrays.asList(this.httpPayload.getQueryMap(), this.httpPayload.getParamsMap());
+        this.httpHeaders = this.httpPayload.getHeadersMap();
```
JunyiYi: Remove the header from `this.fields` is not enough, you also need to make sure the metadata is also resolved by name just like [the one here](https://github.com/pragnagopa/azure-functions-java-worker/blob/6499f80522dd5c5b7ccf085fafc4450d120ceaa1/src/main/java/com/microsoft/azure/functions/binding/RpcComplexDataSources.java#L91).

The method above is used to provide the value when developers are using `@Bingding("header-name")` annotation.

src/main/java/com/microsoft/azure/functions/binding/RpcDataTargets.java
```diff
@@ -24,23 +24,27 @@
     @Override
     public void setStatus(int status) { this.status = status; }
     @Override
-    public void addHeader(String key, String value) { this.headers.put(key, value); }
+    public void addHeader(String key, Object value) { this.headers.put(key, value); }
```
JunyiYi: Make sure you updated the signature of [the interface in java-library](https://github.com/Azure/azure-functions-java-library/blob/dev/src/main/java/com/microsoft/azure/functions/HttpResponseMessage.java#L6-L7).

src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -225,31 +309,50 @@ message RpcLog {
...
   string message = 2;
 }
 
-// TODO - solidify this
+// TODO - solidify this or remove it
 message RpcHttp {
   string method = 1;
   string url = 2; 
-  map<string,string> headers = 3;
+  map<string,TypedData> headers = 3;
```
JunyiYi: Although it is in another repository, but is it reasonable to have a strongly-typed HTTP header? Because by definition, all HTTP headers are String-String pairs.

src/main/java/com/microsoft/azure/functions/binding/RpcComplexDataSources.java
```diff
@@ -112,7 +113,7 @@ private HttpRequestMessageImpl(RpcHttpRequestDataSource parentDataSource, Object
         @Override
         public String getMethod() { return this.parentDataSource.httpPayload.getMethod(); }
         @Override
-        public Map<String, String> getHeaders() { return this.parentDataSource.httpPayload.getHeadersMap(); }
+        public Map<String, TypedData> getHeaders() { return this.parentDataSource.httpPayload.getHeadersMap(); }
```
JunyiYi: Don't do this because you are exposing an internal data structure to the public developers. Try to get all the actual values (like String, int, etc.) and return a Map with those values.

And make sure the return type matches the [signature here](https://github.com/Azure/azure-functions-java-library/pull/10/files#diff-e6d1900cf8bd12c64341fdc11d63cc43R9).

src/main/java/com/microsoft/azure/functions/binding/RpcComplexDataSources.java
pragnagopa: Let me know what is the right way to fix this. Can you point me to documentation?

src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
pragnagopa: Here is a good example and discussion on valid values for http headers: [using-json-string-in-the-http-header](https://stackoverflow.com/questions/9779860/using-json-string-in-the-http-header)

src/main/java/com/microsoft/azure/functions/binding/RpcComplexDataSources.java
pragnagopa: Isn't RpcComplexDataSources.java internal to worker? How is this exposed to user? I have updated the signature to  Map<String, Object> getHeaders() in azure-functions-java-library.

pragnagopa: Also aren't HttpRequestMessage and HttpResponseMessage are internal implementation details of worker? If yes, can we move them here? If they need to be exposed to user, can you please provide an example?

pragnagopa: @JunyiYi -  I think you are right. If the requirement for http headers is to be <string, string>, we should not be changing RpcHttp in proto file. I will look into this and update the PR soon.

src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
pragnagopa: @JunyiYi - I think you are right. If the requirement for headers to be <string, string>, we should not be changing RpcHttp in proto file. Will get back to you on this.

src/main/java/com/microsoft/azure/functions/binding/RpcComplexDataSources.java
JunyiYi: Let's see whether we need to change the HTTP header at all. If not, there is no need to modify this.

src/main/java/com/microsoft/azure/functions/binding/RpcComplexDataSources.java
JunyiYi: `HttpRequestMessage` is a public interface, it contains a method definition [`Map<String, Object> getHeaders()`](https://github.com/Azure/azure-functions-java-library/pull/10/files#diff-e6d1900cf8bd12c64341fdc11d63cc43R34). And you are returning `TypedData` here (which is not correct actually, because it needs to `@Override` the correct types). The scope in Java is just a compilation stage thing, when it is running, user could access the `TypedData` you are returning, and I think it should be an internal data structure (because it is a data structure used in GRPC).

So rather than returning the `TypedData`, a better way is to convert it to Java native types (or even POJO type), and return it to the user.

pragnagopa:
@JunyiYi - Reverted changes to http headers,

src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -108,113 +135,170 @@ message WorkerTerminate {
   google.protobuf.Duration grace_period = 1;
 }
 
+// Host notifies worker of file content change
 message FileChangeEventRequest {
-  // https://msdn.microsoft.com/en-us/library/t6xf43e0(v=vs.110).aspx
+  // Types of File change operations (See link for more info: https://msdn.microsoft.com/en-us/library/t6xf43e0(v=vs.110).aspx)
   enum Type {
-	Unknown = 0;
+	  Unknown = 0;
```
JunyiYi: Space.

src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -108,113 +135,170 @@ message WorkerTerminate {
...
+// Used to encapsulate data which could be a variety of types
 message TypedData {
   oneof data {
     string string = 1;
     string json = 2;
     bytes bytes = 3;
     bytes stream = 4;
     RpcHttp http = 5;
-	sint64 int = 6;
+	  sint64 int = 6;
```
JunyiYi: Space.

pragnagopa: Will format proto file in https://github.com/Azure/azure-functions-language-worker-protobuf/ repo

PR #112, opened 2018-05-30
Implements changes in azure-functions-java-library PR #12

brunoborges:
See this PR: https://github.com/Azure/azure-functions-java-library/pull/12

Implements the new getHttpMethod that returns Enum, and the return type of setStatus, addHeader, and setBody on HttpResponseMessage

Signed-off-by: Bruno Borges <bruno.borges@gmail.com>


src/main/java/com/microsoft/azure/functions/worker/binding/RpcComplexDataSources.java
```diff
@@ -114,6 +114,8 @@ private HttpRequestMessageImpl(RpcHttpRequestDataSource parentDataSource, Object
         @Override
         public String getMethod() { return this.parentDataSource.httpPayload.getMethod(); }
```
JunyiYi: We can remove this since we are still in the BETA stage, we accept breaking changes.

JunyiYi:
Thanks @brunoborges for your contribution. I just left one simple suggestion, and the code LGTM. I will also add @pragnagopa , since now she is the owner of this repository.

src/main/java/com/microsoft/azure/functions/worker/binding/RpcComplexDataSources.java
brunoborges: While it is beta, there are many places where this is being used. One example is the Spring Cloud Function adapter. 

My take is that we should keep the API, add a Builder, and then deprecate those methods when we release 1.0-final

JunyiYi: It doesn't matter. Because we've already introduced so many breaking changes even like changing the package name.

brunoborges: Ok, I will proceed with that cleanup.

brunoborges:
Rebase done!

pragnagopa:
Thanks!
Published azure-functions-java-library 1.0.0-beta-5 to maven
```
<groupId>com.microsoft.azure.functions</groupId>
<artifactId>azure-functions-java-library</artifactId>
 <version>1.0.0-beta-5</version>
```
Please update pom.xml to refer to the latest library. Will merge this PR once we have a green build.

pragnagopa:
@brunoborges - Can you please update the PR? 
@jdneo - Heads up on the breaking changes in azure-functions-library. Once this PR is merged, let's start updating maven plugin and publish to sonatype repo

brunoborges:
Done!

Issue #143, opened 2018-07-12
End2End test status


pragnagopa:
@JanusWu  - Opening this issue to keep track for E2E test status

asavaritayal:
Checked in.

Issue #142, opened 2018-07-12
Add EventGrid E2E test


pragnagopa:
Add EventGrid E2E test

Issue #130, opened 2018-06-15
Add E2E tests to CI


pragnagopa:
Move tests from https://github.com/songkaiape/azure-function-test

pragnagopa:
@songkaiape - Lets use this issue to track automating Java E2E tests

JanusWu:
@pragnagopa I can't create a new branch in this repo, does this need other permission?
`$ git checkout -b dev/v-tanwu/AddE2Etests`
`fatal: cannot lock ref 'refs/heads/dev/v-tanwu/AddE2Etests': 'refs/heads/dev' ex                                              ists; cannot create 'refs/heads/dev/v-tanwu/AddE2Etests'`

Besides, please assign this issue to me. @songkaiape has left team.

pragnagopa:
Please fork and send a PR from your fork. Create a folder EndToEndTests and copy tests there.

JanusWu:
I have created a PR #135 , Please review.

pragnagopa:
@JanusWu - Can you send a new PR?

PR #158, opened 2018-08-13
Add end to end Azure Functions annotation samples

JanusWu:
Tracking issue #130, #142 , #143.


JanusWu:
@pragnagopa - could you review?  :)

JanusWu:
I'm not sure why this PR contains previous commits, maybe I need to re-fork this repo?

pragnagopa:
Can you rebase and squash your commits?

End2End/functest/src/main/java/com/microsoft/azure/functions/endtoend/ServiceBusTopic.java
```diff
@@ -0,0 +1,24 @@
+package com.microsoft.azure.functions.endtoend;
+
+import com.microsoft.azure.functions.ExecutionContext;
+import com.microsoft.azure.functions.annotation.BindingName;
+import com.microsoft.azure.functions.OutputBinding;
+
+public class ServiceBusTopic {
+
+    public static void serviceBusTopicTrigger(
+        @BindingName("message") String message,
```
pragnagopa: Changes looks good. We do need to update all the tests to use annotations and delete checked in function.json files.

JanusWu:
I'm confused. Delete all function.json files and only use annotations, you mean make the test like which in [Azure-Functions-Java-Samples](https://github.com/jdneo/Azure-Functions-Java-Samples/tree/master/ServiceBusQueueTrigger/project)?
But at the very beginning, we decided to move the annotation tests to azure-maven-plugins, and keep the tests with functions left. Right? May I know why we have this change?

Besides, If we do this change, the tests of HTTP, Queue, Timer and Blob which were already merged into java-worker also need to be changed. Right?
Correct me if I'm wrong.

JanusWu:
@pragnagopa How about now?

pragnagopa:
Commits look good :) Thanks! Can you confirm if the all tests in this PR are passing?

JanusWu:
Add Cosmos DB test and squash the commit.

JanusWu:
I have confirmed that all the tests are passing. :)

pragnagopa:
Thanks @JanusWu. I will go ahead and merge them in. Will try out E2E tests as well.
cc: @paulbatum, @selvasingh  - We now have working versions of tests for all the triggers including CosmosDB.

PR #212, opened 2018-10-17
Cleanup

### LINKED ISSUE #205: Remove support for overloading functions
See related discussion here: https://github.com/Azure/azure-functions-java-worker/issues/26
Method overloading sort of is supported but most of the scenarios it throws "NoSuchMethodExceptions"

There is support for this in webjobs sdk as well. Having it here and documented:
As Documented here: https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-java#function-method-overloading adds more complexity on the worker to look up which funciton to excecute.

pragnagopa:
Fixes #205


src/main/java/com/microsoft/azure/functions/worker/broker/OverloadResolver.java
```diff
@@ -69,22 +53,21 @@ else if (overloadComparator.compare(possibleInvokes.get(0), possibleInvokes.get(
...
             }
             if (!method.entry.getReturnType().equals(void.class) && !method.entry.getReturnType().equals(Void.class)) {
                 dataStore.getOrAddDataTarget(invokeInfo.outputsId, BindingDataStore.RETURN_NAME, method.entry.getReturnType());
             }
-            return Optional.of(invokeInfo);
+            return invokeInfo;
         } catch (Exception ex) {
-            return Optional.empty();
+            //TODO log
+            return null;
```
paulbatum: Planning on addressing this todo in thisi PR?

pragnagopa: Still figuring out  which exceptions are useful to log. Will address this separately.

PR #226, opened 2018-10-29
Remove dependency on guava

pragnagopa:


pragnagopa:
@safihamid fyi - removing dependency on netty and guava

pragnagopa:
@safihamid - Looks like we do need netty.

Issue #183, opened 2018-09-18
Investigate dependecy conflicts between worker and user code


pragnagopa:
An example of conflict #182 
Need to investigate how users can bring in their dependencies that conflict with Java worker

asavaritayal:
At the moment, we have the following runtime dependencies for the language worker: 

- azure-functions-java-library
- protobuf-java
- grpc-protobuf
- grpc-stub
- grpc-netty
- guava
- commons-lang3
- commons-cli
- jackson-databind
- jackson-annotations
- javax.annotation-api
- junit

asavaritayal:
From @brunoborges -

Only way I can think of is if you ensure user-defined dependencies are loaded first in the clasapath. But that may break the worker if user chooses an incompatible version of a dependency.

In other words: I don't think it is possible.

As I mentioned couple months ago, right now the architecture has worker and user code tightly coupled.

The ideal solution is to have user code running in one Java process, and the worker in another one.

This is how OpenFaaS and FN Project are designed.

brunoborges:
Just to provide more details: the other option is to isolate user code with a different ClassLoader. But that imposes several other challenges.

pragnagopa:
This issue becomes more important when we start supporting rich types as that would require worker to take dependency on azure libraries for each component

pragnagopa:
Related issue #117 
Reducing worker dependencies on third party libraries and allowing the users to configure classpath might help

Fabiest:
Hello,

Sorry for "re-opening" this.
I'm having a weird error with the jackson lib. I'm trying to use lib A that depends on jackson annotations 2.9.6. When i run my stuff under azure functions, the code fails because it cannot find property USER_DEFAULT from jackson code. Such property only exists after jackson 2.6.0.
Now, i've tried to force 2.9.6 from my pom and reading this thread, it makes no sense that i have this issue (first, it seems func worker was using 2.9.6 and it seems you have removed jsckson dependency to use gson).

What can i do? What is that alternative around the classpath? I'm not really  ajava developer so i'm not sure what i can do.

Fabiest:
Hello,

I'm suffering from similar problem where I want to use lib A which depends on Jackson 2.9.6 (annotations, binding, core) and lib 8 which depends on Jackson 2.4.0 (annotations).

[...]

PR #232, opened 2018-11-06
Swtich to Gson from Jackson

pragnagopa:


pragnagopa:
@safihamid - Can we please try bits from my branch to check if this improves cold start?

pragnagopa:
@fabiocav / @paulbatum  - FYI if this does help, we should plan for a runtime release and it would remove dependency on Jackson but add dependency on GSON

paulbatum:
We should be careful as this might be a breaking change (e.g. what happens if a java function app is referencing GSON when we make this change).

pragnagopa:
> We should be careful as this might be a breaking change (e.g. what happens if a java function app is referencing GSON when we make this change).

Yes, this would be a breaking change as this would enforce the Gson version dependency. Users will not be able to bring their own version.

pragnagopa:
Just realized gson is already included in the worker as part of protobuf dependency
```
[INFO] Including com.google.code.gson:gson:jar:2.8.5 in the shaded jar.
```
This change is an improvement as it removes dependency on Jackson.

pragnagopa:
Also, confirmed this is not a breaking change for apps using Jackson as they had to explicitly bring in that dependency as part of the function code.

nicenemo:
Nice to know, why Gson and not...?
(Have to make that choice in another project)

We are using Gson in Azure functions together with storage queues, blobs and storage tables. What will break?

Didn't you have Gson as an indirect dependency? I remember having to add a later version to my pom.xml because Snyk complained about vulnerabilities.

Trosky:
@pragnagopa have you tried to deserialize a value just with a jackson annotations dependency present in the function? 

We're making heavy use of this "hidden" feature:
Example:
```

    private class Order {

        private String something;
        @com.fasterxml.jackson.annotation.JsonIgnore
        private String userId;

        public String getSomething() {
            return something;
        }

        public void setSomething(String something) {
            this.something = something;
        }

        public String getUserId() {
            return userId;
        }

        public void setUserId(String userId) {
            this.userId = userId;
        }
    }

    @FunctionName("userOrders")
    public HttpResponseMessage getUserOrders(
            @HttpTrigger(name = "req",
                    methods = {HttpMethod.GET},
                    authLevel = AuthorizationLevel.ANONYMOUS,
                    route = "users/{userId}/orders")
                    HttpRequestMessage<Optional<?>> orderRequest,
            @BindingName("userId") String userId,
        context.getLogger().info("OrderFunction!");
        Order order = new Order();
        order.setSomething("something");
        order.setUserId(userId);

        return orderRequest.createResponseBuilder(HttpStatus.OK).body(order).build();

    }

```

returns:
```
 {
    "something": "something"
  }
```

pragnagopa:
@nicenemo 
> Nice to know, why Gson and not...?

From functionality perspective both Jackson and Gson are identical. Main motivation behind this switch is to minimize the dependencies brought in by azure-functions-java-worker as we are yet to solve https://github.com/Azure/azure-functions-java-worker/issues/183.  azure-functions-java-worker relies on grpc and protobuff which already have a dependency on gson. This change is to leverage existing dependency instead adding another library (jackson) for the same purpose

> We are using Gson in Azure functions together with storage queues, blobs and storage tables. What will break?

No breaking changes with this change

> Didn't you have Gson as an indirect dependency? I remember having to add a later version to my pom.xml because Snyk complained about vulnerabilities.

Yes, GSON is still an indirect dependency brought in by the azure-functions-java-worker at runtime. As dependencies from the azure-functions-java-worker are loaded in the class path first, you will be locked into the version gson v 2.8.5 . We will continue to investigate #183 but we do not have an ETA

pragnagopa:
@Trosky 
I am assuming your azure function's pom.xml has a dependency on com.fasterxml.jackson which will continue to work as expected. With this change you are not locked into version that azure-functions-java-worker used to depend on.

Trosky:
@pragnagopa 
We only have the following dependency to Jackson:
```
        <dependency>
            <groupId>com.fasterxml.jackson.core</groupId>
            <artifactId>jackson-annotations</artifactId>
        </dependency>
```

As far as I know GSON does not respect the annotations from Jackson.
The following Test would fail:
```
public class SomethingTest {
    private class Order {

        private String something;
        @com.fasterxml.jackson.annotation.JsonIgnore
        private String userId;

        public String getSomething() {
            return something;
        }

        public void setSomething(String something) {
            this.something = something;
        }

        public String getUserId() {
            return userId;
        }

        public void setUserId(String userId) {
            this.userId = userId;
        }
    }

    @Test
    public void testSomething() {
        Gson gson = new Gson();
        Order order=new Order();
        order.setSomething("something");
        order.setUserId(UUID.randomUUID().toString());

        assertEquals("{\"something\":\"something\"}",gson.toJson(order));
    }
}
```

pragnagopa:
@Trosky  - Please open a separate issue with details including your function code.

pragnagopa:
@bernerha - To have greater control over serialization, function code should not rely on java worker's behavior. Please see sample code here https://github.com/Azure/azure-functions-java-worker/issues/236 for the recommended approach - this gives the ability to bring in your own version of jackson instead of being locked to a version provided by the runtime.

bernerha:
Ok. So you are going to remove the automatic serialization at all on the responses? Jackson and Gson?

Trosky:
I agree with @bernerha. If the worker already has a serialization framework, we should also use it to have less dependencies. 

Why is that not recommended? You could even make the mapper explicitly configurable by the Users.

TurtleBeach:
I am thoroughly frustrated with this esoteric back and forth about Gson vs Jackson, while in the meantime ALL of my Java functions IN PRODUCTION are now broken because GSON uses a Google defined **INTERNAL** class called a LinkedTreeMap that does not cast to a standard Java HashMap (which previously would work with a JSON body in a POST request).
What is even more frustrating is that the latest (as of November 13) POM that is constructed when a new Java function App is created does NOT have the dependencies for Gson. All of the links I have found in this thread or related threads point to Gson 2.8.5, but Maven doesn't find that version, I think because of the ...SONATYPE repository references that are placed in the POM.
I really cannot fathom why you folks at Microsoft seemingly completely disregard people who are actually trying to create something by blithely introducing breaking changes for trivial reasons. Come on - eliminating one dependency (which if your systems are automated means absolutely nothing other than perhaps some trivial amount of wasted storage somewhere), with the consequence of breaking code that has been running (Preview for JAVA excuses be damned - Functions are touted as GA, and Java has been around for over a year). MS did the same thing with changing an int to an enum for absolutely no good reason (in my humble judgment) - but at least that happened before V2 functions were touted as GA.
If you are going to make these kinds of changes then at least have the professional courtesy to document clearly what is required and make sure the tools are ready before you role out the changes.
In short, what should a POM now look like to get the required Gson????
Jack Copper

TurtleBeach:
Further to my comments above - after wasting a ridiculous amount of time fixing the POM from the azure function prototype, I again have to wonder why MS deemed it necessary to implement a JSON library that relies on undocumented private Google data structures (the JavaDoc for Gson 2.8.5 has NO reference to a class "LinkedTreeMap", which is what the debuggers (Visual Studio Code and Netbeans) report is the class of internal objects which, by JSON standards, should simply be Maps (or HashMaps). To Wit, this simple structure:
{
   "User":
   {
      "UserGUID":"a1b2f692-cd76-4fc9-80c9-26fe77e3d33a"
   }
}
which is sent as application/json payload (body) to a function, is a HashMap (as expected) when retrieved as request.getBody() inside the function. BUT the inner object ("user" above) which previously was also a HashMap, is now an undocumented com.google.gson.internal.LinkedTreeMap which cannot be cast to a HashMap, a LinkedHashMap, or a Map. Sure, it has methods like a HashMap, but if that is the case WHY NOT USE A STANDARD HASHMAP???. I carefully designed communication structures based on standard Java classes, and the structures have worked for over a year. Now MS demands I change source code to use an undocumented non-standard class that for all I know will go away the next time someone at MS decides to use the next cute thing because they can't figure out how to fix their own architecture.
Unbelievable.
If I wanted to inextricably tie my code to Google trinkets I would not be using Azure as a platform.
As you can infer, I am not a happy camper. You guys and gals need to get your collective acts together.
Jack Copper

TurtleBeach:
In addition, the Gson parser is brain-dead with respect to integers - so everywhere I relied on casting values to Integers because I knew integers were sent (numbers with no decimal point), and which the previous parser returned as an Integers, I now have to get the intValue of a Double since Gson chooses to return everything as a Double. A debate about the assumptions made by a parser vis a vis strings representing numbers that do not contain a decimal separator might have been fair game a year ago, but not after code has been out in the wild for over a year. So yet more slogging through sources to change accommodate change to fundamental behavior that was previously reliable and working.

pragnagopa:
This PR is now merged and closed.
@Trosky  - Exposing serializer settings is a feature request. Please open a separate issue.
@TurtleBeach - If you need further assistance, open a separate issue with concise description. Provide a repro, expected and actual behavior.

UltimaPhoenix:
@pragnagopa 
> From functionality perspective both Jackson and Gson are identical. 

IMHO The number of features in Jackson are far more than in Gson that's why I consider Jackson better. For example, I didn't find any annotation like `@JsonAnySetter` and `@JsonAnyGetter` in Gson and they are fundamental to build a future proof application (which allows to read and propagate all fields in a pipeline).

> Main motivation behind this switch is to minimize the dependencies brought in by azure-functions-java-worker as we are yet to solve #183. 

This is not a true solution to the dependency management. Manage multiple dependencies is something that must be done for every project, swapping a library with another one just moves the issue to users that use a different Gson version instead of Jackson.  BTW it's true that reducing the number of dependency is a good way to minimize risks so Gson can be a good choice but not for my use cases.

Can leaving the choice between Jackson/Gson to the user be considered in the next releases?

MANDA-LAWINE:
What a bad feature this is. Should be reverted IMO.

For everyone having a bad day with serialization/deserialization in azure function, this is the root cause :D

PR #248, opened 2018-11-27
Add support to update environment variables

pragnagopa:
Adds support to reload environment variables

### DIFF
--- a/src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
+++ b/src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
@@ -5,6 +5,7 @@ option java_multiple_files = true;
 option java_package = "com.microsoft.azure.functions.rpc.messages";
 option java_outer_classname = "FunctionProto";
 option csharp_namespace = "Microsoft.Azure.WebJobs.Script.Grpc.Messages";
+option go_package ="github.com/Azure/azure-functions-go-worker/internal/rpc";
 
 package AzureFunctionsRpcMessages;
 
@@ -64,6 +65,10 @@ message StreamingMessage {
 
     // Worker logs a message back to the host
     RpcLog rpc_log = 2;
+
+    FunctionEnvironmentReloadRequest function_environment_reload_request = 25;
+
+    FunctionEnvironmentReloadResponse function_environment_reload_response = 26;
   }
 }
 
@@ -177,6 +182,16 @@ message WorkerStatusRequest{
 
 // NOT USED
 message WorkerStatusResponse {
+}
+
+message FunctionEnvironmentReloadRequest {
+  // Environment variables from the current process
+  map<string, string> environment_variables = 1;
+}
+
+message FunctionEnvironmentReloadResponse {
+  // Status of the response
+  StatusResult result = 3;
 }
 
 // Host tells the worker to load a Function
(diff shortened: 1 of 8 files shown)


src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
```diff
@@ -0,0 +1,73 @@
...
+    return String.format("FunctionEnvironmentReloadRequest completed");
+  }
+
+  FunctionMethodDescriptor createFunctionDescriptor(String functionId,
+      RpcFunctionMetadata metadata) {
+    return new FunctionMethodDescriptor(functionId, metadata.getName(), metadata.getEntryPoint(),
+        metadata.getScriptFile());
+  }
+  
+  public static void setEnv(Map<String, String> newSettings) throws Exception {
```
pragnagopa: @fabiocav - Java does not have a built in helper to set the environment variables. Following code snippet is taken from [SO post](https://stackoverflow.com/questions/318239/how-do-i-set-environment-variables-from-java)

src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
```diff
@@ -5,6 +5,7 @@ option java_multiple_files = true;
 option java_package = "com.microsoft.azure.functions.rpc.messages";
 option java_outer_classname = "FunctionProto";
 option csharp_namespace = "Microsoft.Azure.WebJobs.Script.Grpc.Messages";
+option go_package ="github.com/Azure/azure-functions-go-worker/internal/rpc";
```
fabiocav: Can you help me understand what this is for?

src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -122,7 +123,7 @@ public static void main(String[] args) {
             try (JavaWorkerClient client = new JavaWorkerClient(app)) {
                 client.listen(app.getWorkerId(), app.getRequestId()).get();
             } catch (Exception ex) {
-                WorkerLogManager.getSystemLogger().log(Level.SEVERE, "Unexpected Exception causes system to exit", ex);
+                WorkerLogManager.getSystemLogger().log(Level.SEVERE, ExceptionUtils.getRootCauseMessage(ex), ex);
```
fabiocav: It this an unrelated change?

src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
fabiocav: Wow... this looks pretty fragile. Should we have a note here?

src/main/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto
pragnagopa: as of now go support is in private preview.

src/main/java/com/microsoft/azure/functions/worker/Application.java
pragnagopa: Yes. Improving error logging to help us get the exact exception message

PR #253, opened 2018-12-14
Adding PicoHelper.Java

safihamid:


src/main/java/com/microsoft/azure/functions/worker/handler/PicoHelper.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.sun.jna.win32.StdCallLibrary;
+import com.sun.jna.Native;
+import com.sun.jna.win32.W32APIOptions;
+
+// kernel32.dll uses the __stdcall calling convention (check the function
+// declaration for "WINAPI" or "PASCAL"), so extend StdCallLibrary
+// Most C libraries will just extend com.sun.jna.Library,
+public interface PicoHelper extends StdCallLibrary { 
+    PicoHelper INSTANCE = (PicoHelper)Native.load("picohelper", PicoHelper.class, W32APIOptions.DEFAULT_OPTIONS);
```
pragnagopa: @safihamid - Pushed changes to fix build errors. Unit test is failing with error
```
java.lang.UnsatisfiedLinkError: Unable to load library 'picohelper': Native library (win32-x86-64/picohelper.dll) not found in resource path 
```
missed adding picphelper.dll?

pragnagopa: Thanks. Will update the test.

src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
```diff
@@ -43,36 +50,49 @@ FunctionMethodDescriptor createFunctionDescriptor(String functionId,
...
-	      }
-	    }
-	  }
-	}
 
+    for (Map.Entry<String,String> entry : newSettings.entrySet()) {
+        Kernel32.INSTANCE.SetEnvironmentVariable(entry.getKey(), entry.getValue());
+      }
+
+      Kernel32.INSTANCE.SetEnvironmentVariable("WEBSITE_PLACEHOLDER_MODE", null);
```
pragnagopa: two calls required? Can't we just set `WEBSITE_PLACEHOLDER_MODE` to 0 directly?

safihamid: Basically we will need to do a merge here. meaning delete the env variables which don't exist anymore by setting them to null. update the existing ones and add the new ones. so this part needs to be updated.

src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
```diff
@@ -39,40 +46,53 @@ FunctionMethodDescriptor createFunctionDescriptor(String functionId,
...
-	        Map<String, String> map = (Map<String, String>) obj;
-	        map.clear();
-	        map.putAll(newSettings);
-	      }
-	    }
-	  }
-	}
 
+    // This is required to reinitialize file detours
+    newSettings.put(WebsitePlaceholderMode, "0");
```
safihamid: Please note we will have to call Kernel32.INSTANCE.SetEnvironmentVariable(entry.getKey(), entry.getValue()); for all these environment variables so the process env variable is updated not only JVM env variable.
What you updated here only updates what's visible to JVM. That's why I put the code there to call  Kernel32.INSTANCE.SetEnvironmentVariable

pragnagopa: > for all these environment variables so the process env variable is updated not only JVM env variable.
Can you please clarify process vs JVM? are you referring to system environment variables? they would be set by the parent - functions host process right? I am trying to establish a pattern for specialization for other language workers as well.

JonathanGiles:
I'm not familiar with PicoHelper, so I'm interested in learning why this platform-specific dependency is being added. I would be nervous that this would become a roadblock at a later date should functions run on Linux, for example?

src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
safihamid: We used ProcessExplorer and noticed with the existing code, not all env variables were updated. so that's why added the new lines. If you can verify using ProcessExplorer that whatever is sent is added/updated and not existing ones get deleted then we should be fine

pragnagopa:
@JonathanGiles - This PR is to address cold start issue - https://github.com/Azure/azure-functions-host/issues/3872 specifically when running on windows. We are working on a feature to warm up java language worker in order to improve cold start. I will document the design once we have ironed out all the issues.

> I would be nervous that this would become a roadblock at a later date should functions run on Linux, for example?

Yes. When we do add support to run java functions on Linux, we need to update the specialization code. But note that this is not a blocker of Linux. Work involved is specifically for warming up java worker when running on  Linux

src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
pragnagopa: Thanks! Existing code was intentionally overriding system environment variables with the new dictionary sent by the functions host. Will update the code to only add new keys and update existing keys

safihamid: And please don't forget to delete the ones which no longer exist.

pragnagopa: Existing code is already doing this. Deletes existing system environment variables and replaces them with the new dictionary sent by the functions host.

> We used ProcessExplorer and noticed with the existing code, not all env variables were updated. 

It is possible that environment variables sent by the functions host are updated only after specialization completes. For example `WEBSITE_PLACEHOLDER_MODE`  is updated by the host after specialization completes but the call to specialize language worker happens before. We should explicitly set these in the language worker.

safihamid: Ok I remember the site name was also still the mawsplaceholder.... instead of actual site name when looking at env variables. but when I added new code it got fixed to the actual site name. we can test that as well.

PR #271, opened 2019-02-13
name from custom annotation overrides

pragnagopa:
If Custom annotation defined name(), it overrides name defined in `com.microsoft.azure.functions.annotation.CustomBinding`

```java
@Target(ElementType.PARAMETER)
@Retention(RetentionPolicy.RUNTIME)
@CustomBinding(direction = "in", name = "", type = "customBinding")
public @interface TestCustomBinding {
   String index();
   String path();
   String name();
}

```


pragnagopa:
@anthonychu  - Updating the worker to give more control to the custom binding author on the name.

anthonychu:
Thanks for this. This is great.

PR #277, opened 2019-03-02
Update to Version to 1.3.0

### LINKED ISSUE #275: Allow EventData[] POJO for Event Hubs batch
A few older issues that are closed now that *may* be related, but to avoid confusion just capturing this ask.  

If you set cardinality to many, you can only today use `string[]` as the type of the trigger, which sends an array of the Event Hub messages (`EventData.Body`).  However that means you lose a lot of important information on the event, like the system and user properties.

I believe there should be a way to preserve the binding properties and send in a POJO without having to go all the way to "live objects."

pragnagopa:
Udpate event hub test to bind to systemproperties array

Fixes #275

### DIFF
--- a/endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/EventHubTriggerTests.java
+++ b/endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/EventHubTriggerTests.java
@@ -24,11 +24,14 @@ public class EventHubTriggerTests {
     @FunctionName("EventHubTriggerAndOutputString")
     public void EventHubTriggerAndOutputString(
         @EventHubTrigger(name = "messages", eventHubName = "test-input-java", connection = "AzureWebJobsEventHubSender", dataType = "string", cardinality = Cardinality.MANY) String[] messages,
+        @BindingName("SystemPropertiesArray") SystemProperty[] systemPropertiesArray,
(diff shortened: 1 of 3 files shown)


endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/EventHubTriggerTests.java
```diff
@@ -24,11 +24,14 @@ public void EventHubTriggerAndOutputJSON(
     @FunctionName("EventHubTriggerAndOutputString")
     public void EventHubTriggerAndOutputString(
         @EventHubTrigger(name = "messages", eventHubName = "test-input-java", connection = "AzureWebJobsEventHubSender", dataType = "string", cardinality = Cardinality.MANY) String[] messages,
+        @BindingName("SystemPropertiesArray") SystemProperty[] systemPropertiesArray,
```
pragnagopa: @jeffhollan - this is the sample code to bind to systemproperties array.

jeffhollan:
This is good to know that it works and definitely drops the priority of the `EventData[]`, but #275 still worth doing just from a quality of life point of view.  If I'm accessing messages I am limited to only using the array index instead of a `foreach` loop since a single EventData and all properties (messageId, systemProperties, userProperties) are all in their own array.  

So this is super helpful to know, but I think #275 should still be tracked longer term (maybe only support when we do live objects)

Thanks @pragnagopa !

pragnagopa:
Having rich type support that is binding to eventData from event hubs sdk is the right next step

PR #313, opened 2019-07-24
Porting E2E test framework from nodejs worker repo

pragnagopa:


pragnagopa:
Note: This is E2E test framework improvement port from nodejs worker repo

PR #360, opened 2020-04-22
Added capability to send status of a worker.

yojagad:


src/main/java/com/microsoft/azure/functions/worker/handler/WorkerStatusRequestHandler.java
```diff
@@ -0,0 +1,17 @@
...
+    public WorkerStatusRequestHandler() {
+        super(StreamingMessage::getWorkerStatusRequest,
+              WorkerStatusResponse::newBuilder,
+              null,
+              StreamingMessage.Builder::setWorkerStatusResponse);
+    }
+
+    @Override
+    String execute(WorkerStatusRequest request, WorkerStatusResponse.Builder response) {
+        return "";
```
mathewc, line 15: Can we write a completed message like the other handlers do, e.g. `return String.format("WorkerStatusRequest completed");`?

amamounelsayed:
@yojagad can you please check the tests?

src/main/java/com/microsoft/azure/functions/worker/handler/WorkerStatusRequestHandler.java
```diff
@@ -0,0 +1,17 @@
...
+
+    public WorkerStatusRequestHandler() {
+        super(StreamingMessage::getWorkerStatusRequest,
+              WorkerStatusResponse::newBuilder,
+              null,
+              StreamingMessage.Builder::setWorkerStatusResponse);
+    }
+
+    @Override
+    String execute(WorkerStatusRequest request, WorkerStatusResponse.Builder response) {
```
amamounelsayed, line 14: Please correct me if I am wrong, this message will not matter as the WorkerStatusResponse is empty in the protobuf message.

yojagad: True, but just leaving it there for future too when we actually update it with contents.

yojagad:
@amamounelsayed Just squashed the commits. Please merge it whenever you can , thank you!

PR #361, opened 2020-04-23
Add support for java11.

amamounelsayed:
This PR is for inital support for supporting Java 11 runtime, this PR includes.

- Add support for java11 testing pipeline.
- Fix the dev pipeline.
- Add support for java11.

Next, 
- Investigating updating classloader to lad last first, this should fix the dependency issue.
- Investigating the support for Modules.
- Support the two separate pipelines for java11 and java8.

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/broker/FactoryJavaMethodExecutor.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/broker/FactoryJavaMethodExecutor.java
@@ -1,0 +1,23 @@
+package com.microsoft.azure.functions.worker.broker;
+
+import com.microsoft.azure.functions.rpc.messages.BindingInfo;
+import com.microsoft.azure.functions.worker.WorkerLogManager;
+import com.microsoft.azure.functions.worker.description.FunctionMethodDescriptor;
+import com.microsoft.azure.functions.worker.reflect.ClassLoaderProvider;
+import org.apache.commons.lang3.SystemUtils;
+
+import java.net.MalformedURLException;
+import java.util.Map;
+
+public class FactoryJavaMethodExecutor {
+    public JavaMethodExecutor getJavaMethodExecutor(FunctionMethodDescriptor descriptor, Map<String, BindingInfo> bindings, ClassLoaderProvider classLoaderProvider)
+            throws MalformedURLException, ClassNotFoundException, NoSuchMethodException {
+        if(SystemUtils.IS_JAVA_1_8) {
+            WorkerLogManager.getSystemLogger().info("Loading JavaMethodExecutorImpl");
+            return new JavaMethodExecutorImpl(descriptor, bindings, classLoaderProvider);
+        } else {
+            WorkerLogManager.getSystemLogger().info("Loading EnhancedJavaMethodExecutorImpl");
+            return new EnhancedJavaMethodExecutorImpl(descriptor, bindings, classLoaderProvider);
+        }
+    }
+}
--- a/src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
@@ -1,0 +1,76 @@
+package com.microsoft.azure.functions.worker.reflect;
+
+import java.io.*;
+import java.net.*;
+import java.sql.Driver;
+import java.util.*;
+import java.util.concurrent.*;
+
+
+import com.microsoft.azure.functions.worker.*;
+
+public class EnhancedClassLoaderProvider implements ClassLoaderProvider {
+    public EnhancedClassLoaderProvider() {
+        urls = Collections.newSetFromMap(new ConcurrentHashMap<URL, Boolean>());
+    }
+
+    /*
+     * @see com.microsoft.azure.functions.reflect.ClassLoaderProvider#createClassLoader()
+     */
+    @Override
+    public ClassLoader createClassLoader() {
+        URL[] urlsForClassLoader = new URL[urls.size()];
+        urls.toArray(urlsForClassLoader);
+
+        URLClassLoader classLoader = new URLClassLoader(urlsForClassLoader);
+        loadDrivers(classLoader);
+        return classLoader;
+    }
+
+    private void loadDrivers(URLClassLoader classLoader) {
+        Thread.currentThread().setContextClassLoader(classLoader);
+        try {
+            ServiceLoader<Driver> loadedDrivers = ServiceLoader.load(Driver.class);
+            Iterator<Driver> driversIterator = loadedDrivers.iterator();
+            try {
+                while (driversIterator.hasNext()) {
+                    driversIterator.next();
+                }
+            } catch (Throwable t) {
+                // Do nothing
+            }
+        } finally {
+            Thread.currentThread().setContextClassLoader(ClassLoader.getSystemClassLoader());
+        }
+    }
+
+    @Override
+    public void addDirectory(File directory) throws MalformedURLException, IOException {
+        if (!directory.exists()) {
+            return;
+        }
+
+        File[] jarFiles = directory.listFiles(new FileFilter() {
+            @Override
+            public boolean accept(File file) {
+                return file.isFile() && file.getName().endsWith(".jar");
+            }
+        });
+
+        for (File file : jarFiles) {
+            addUrl(file.toURI().toURL());
+        }
+    }
+
+    @Override
+    public void addUrl(URL url) throws IOException {
+        if (urls.contains(url)) {
+            return;
+        }
+
+        WorkerLogManager.getSystemLogger().info("Loading file URL: " + url);
+
+        urls.add(url);
+    }
+    private final Set<URL> urls;
+}
--- a/src/main/java/com/microsoft/azure/functions/worker/reflect/FactoryClassLoader.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/reflect/FactoryClassLoader.java
@@ -1,0 +1,13 @@
+package com.microsoft.azure.functions.worker.reflect;
+
+import org.apache.commons.lang3.SystemUtils;
+
+public class FactoryClassLoader {
+    public ClassLoaderProvider createClassLoaderProvider(){
+        if(SystemUtils.IS_JAVA_1_8) {
+            return new DefaultClassLoaderProvider();
+        } else {
+            return new EnhancedClassLoaderProvider();
+        }
+    }
+}
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,7 +11,7 @@ steps:
    checkLatest: true
 - pwsh: |
      Get-Command mvn
-  displayName: 'Installing Maven'
+  displayName: 'Check Maven is installed'
 - pwsh: |
       $buildNumber = 0      
       if($env:APPVEYOR_REPO_TAG -eq "true") {
@@ -32,10 +32,10 @@ steps:
     CleanTargetFolder: true
   displayName: 'Copying files for artifacts'
 - pwsh: |
-      .\setup-tests.ps1
+      .\setup-tests-pipeline.ps1
   displayName: 'Setting tests'
 - pwsh: |
-      .\run-tests.ps1
+      .\build-run-tests-pipeline.ps1
   env:
     AzureWebJobsStorage: $(AzureWebJobsStorage)
     AzureWebJobsCosmosDBConnectionString: $(AzureWebJobsCosmosDBConnectionString)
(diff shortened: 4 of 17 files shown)


src/main/java/com/microsoft/azure/functions/worker/broker/FactoryJavaMethodExecutor.java
```diff
@@ -0,0 +1,23 @@
...
+import com.microsoft.azure.functions.worker.reflect.ClassLoaderProvider;
+import org.apache.commons.lang3.SystemUtils;
+
+import java.net.MalformedURLException;
+import java.util.Map;
+
+public class FactoryJavaMethodExecutor {
+    public JavaMethodExecutor getJavaMethodExecutor(FunctionMethodDescriptor descriptor, Map<String, BindingInfo> bindings, ClassLoaderProvider classLoaderProvider)
+            throws MalformedURLException, ClassNotFoundException, NoSuchMethodException {
+        if(SystemUtils.IS_JAVA_1_8) {
```
mnriem, line 15: Not needed to use SystemUtils here. You can replace it with a try catch (catching Throwable) using

boolean isJava8 = true;
try {
  Runtime.Version.feature()
  isJava8 = false;
}
catch(Throwable t) {
}

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -0,0 +1,76 @@
...
+    /*
+     * @see com.microsoft.azure.functions.reflect.ClassLoaderProvider#getClassLoader()
+     */
+    @Override
+    public ClassLoader getClassLoader() {
+        URL[] urlsForClassLoader = new URL[urls.size()];
+        urls.toArray(urlsForClassLoader);
+
+        URLClassLoader classLoader = new URLClassLoader(urlsForClassLoader);
+        loadDrivers(classLoader);
```
mnriem, line 26: You are loading the drivers here as a side effect of calling getClassLoader. I would recommend where you call getClassLoader to be followed by the code from loadDrivers

pragnagopa:
@amamounelsayed - Thanks for the PR. Few initial questions:
- Can you point me to build script changes that produce worker jar for Java8 and Java 11?
- Are E2E tests run against Java8 and Java 11
- Worker.config.json updates to figure out worker path for Java 8 and Java 11?

amamounelsayed:
@pragnagopa Thank you so much, The changes were not major to have two workers. So the plan is to have the same worker.
- Can you point me to build script changes that produce worker jar for Java8 and Java 11?
It has the same worker so no need to distinguish between both.
- Are E2E tests run against Java8 and Java 11
Yes the pipeline now have a section for java11 and run the same set of the E2E tests.
- Worker.config.json updates to figure out worker path for Java 8 and Java 11? The same worker need to have it.

Background, to insure that this will affects any java8 users, @kevinhillinger suggested to create Factories that will hide this functionality based on the Java runtime process. Current users should not get any impact.

The code will be compiled by java 8 bites which is compatible with java 11 runtime.

The direction built for java 11 should work for java 8. So to not impact java 8 users we will are going to test it with java 11 and then future plans to sunset the current way and be java8 and 11 work in the same way, but this long term plan.

src/main/java/com/microsoft/azure/functions/worker/broker/FactoryJavaMethodExecutor.java
amamounelsayed: I believe Runtime.Version is introduced in java 9, we are still code compatible with java 8. https://stackoverflow.com/questions/2591083/getting-java-version-at-runtime

amamounelsayed: The reason I used SystemUtils.IS_JAVA_1_8, it is more cleaner to hide the functionality, as it is calling at the end java.specification.version

mnriem: I mentioned this here in case you would want to get rid of the use of SystemUtils. Note this suggestion will work on both versions (it does require Java 9+ to do the compile). Maybe an simple opportunity target if 2 branches are going to be maintained.

pragnagopa:
> The code will be compiled by java 8 bites which is compatible with java 11 runtime.

Apologies if I misunderstood. This PR is making the code change so that worker jars can be produced for Java 8 and Java 11 right? Or are you planning to ship Java 8 worker for Java 11 apps? If later, is there any advantage of doing it this way? As Java 11 is already available in the Platform, for me it makes more sense to create explicit binaries that will also allow evolving Java worker to take advantage to new feature introduced in Java 11

amamounelsayed:
> Apologies if I misunderstood. This PR is making the code change so that worker jars can be produced for Java 8 and Java 11 right? Or are you planning to ship Java 8 worker for Java 11 apps? If later, is there any advantage of doing it this way? As Java 11 is already available in the Platform, for me it makes more sense to create explicit binaries that will also allow evolving Java worker to take advantage to new feature introduced in Java 11

Yes this is great point. Your second assumption is correct, I will add that java8 worker now support java11. Our initial thought is to have two separate worker but currently we don't have a need to adopt any java 11 new features. 
Plus the support for two different branches for any new feature and maintaining it is a burden we do not currently need it.

Once we have a need to adopt new features then we will for sure follow each branch route.

pragnagopa:
> Plus the support for two different branches for any new feature and maintaining it is a burden we do not currently need it.

You do not have to create two branches until there is a need to diverge. Right now, you just have to produce binaries for each version. Once you have the build setup and tests running for both versions, it will allow you to expand in future and create branches if there is a need.

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
amamounelsayed: I updated the name as we discussed to createClassLoader.

amamounelsayed:
> > Plus the support for two different branches for any new feature and maintaining it is a burden we do not currently need it.
> 
> You do not have to create two branches until there is a need to diverge. Right now, you just have to produce binaries for each version. Once you have the build setup and tests running for both versions, it will allow you to expand in future and create branches if there is a need.

Thank you so much for your help. Yes we can for sure work on this direction I created this issue to track it https://github.com/Azure/azure-functions-java-worker/issues/362. For this release we may use the same java worker as there should not be impact to enhance it in future.

pragnagopa:
> are you planning to ship Java 8 worker for Java 11 apps?

This is a question as I do not have expertise on how JVM handles loading references. Can you please confirm following works:
- Java8 worker with Java 11 function app
- Function App is built for Java 11 only i.e it uses new api etc
- Function App has references to libraries that are built Java 11 only.

You need to verify , when launching Java8 worker in Java 11 JVM, it allows user apps to load Java 11 references.

src/main/java/com/microsoft/azure/functions/worker/broker/FactoryJavaMethodExecutor.java
```diff
@@ -0,0 +1,23 @@
...
+import com.microsoft.azure.functions.worker.WorkerLogManager;
+import com.microsoft.azure.functions.worker.description.FunctionMethodDescriptor;
+import com.microsoft.azure.functions.worker.reflect.ClassLoaderProvider;
+import org.apache.commons.lang3.SystemUtils;
+
+import java.net.MalformedURLException;
+import java.util.Map;
+
+public class FactoryJavaMethodExecutor {
+    public JavaMethodExecutor getJavaMethodExecutor(FunctionMethodDescriptor descriptor, Map<String, BindingInfo> bindings, ClassLoaderProvider classLoaderProvider)
```
trask, line 13: I think this would be nice as a static `create(...)` method on the `JavaMethodExecutor` interface (then no need for this Factory class)

src/main/java/com/microsoft/azure/functions/worker/reflect/FactoryClassLoader.java
```diff
@@ -0,0 +1,13 @@
+package com.microsoft.azure.functions.worker.reflect;
+
+import org.apache.commons.lang3.SystemUtils;
+
+public class FactoryClassLoader {
+    public ClassLoaderProvider createClassLoaderProvider(){
```
trask, line 6: same as above, I think this would be nice as a static `create()` method on the `ClassLoaderProvider` interface (then no need for this Factory class)

pragnagopa:
@amamounelsayed  -  For Java worker dependencies are listed here: https://github.com/Azure/azure-functions-java-worker/wiki/Azure-Java-Functions-Worker-Dependencies,  When a function app loads and has overlapping dependencies on any of the dependencies listed above, library from java worker take precedence i.e
 -  If any of the libraries are updated for Java 11, it is possible that users will see not see expected behavior when running in Azure Functions.

Addressing: https://github.com/Azure/azure-functions-java-worker/issues/362 will avoid issues with loading libraries. We already have established pattern for managing language worker versions. Please sync up with @AnatoliB  if you need guidance as power shell worker recently added support for PS7.

AnatoliB:
> * If any of the libraries are updated for Java 11, it is possible that users will see not see expected behavior when running in Azure Functions.

@amamounelsayed and I discussed this issue and the way versioning for PS6/7 workers is implemented. My recommendation was to avoid the overhead and proceed with a single worker _if this is acceptable for the language runtime_. This condition is _very_ important. In PowerShell, there is no practical way to have a single worker handling both PS6 and PS7, and this is why we separated them, but this separation is not free. I don’t know enough about the Java worker to agree or disagree that the Java worker case is different, but I understand the potential concerns raised by @pragnagopa, so these questions need to be answered first.

mnriem:
@AnatoliB For the time being it can be the same as Java11 can run Java8 byte code without any issues, but I foresee that our community will start asking for features (e.g. features geared towards the module system) that can only be delivered by a Java 11 specific runtime.

src/main/java/com/microsoft/azure/functions/worker/broker/FactoryJavaMethodExecutor.java
amamounelsayed: Thank you so much @trask, Just wondering is it common to have the interface has reference to impl classes? I mean if I moved to the interface I will need to import both impl. So if I delete any impl I need to update the interface.

trask: Static methods aren't part of the interface contract, so just a way to organize static helper methods.

I think it makes for a nicer/more discoverable API in a case like this, since you typically know the interface that you want to instantiate, and now you don't need go find/remember the name of a separate factory class to do the instantiation.

One example where I've seen this used to good effect recently: https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/util/context/Context.java

But definitely no worries if you keep it the way it is.

amamounelsayed:
Thank you so much @mnriem and @AnatoliB. @AnatoliB, as @mnriem mentioned it is the case that single worker can handle both java 8 and 11 using java8 bytecode.

@pragnagopa There is two parts in your question compile and runtime.  Any of the jars mentioned it may be complied using any java version and its behavior is guarantee. For run time it should run anywhere and newer runtime version will run it for backward compatible. 

For testing on webapp yes it has been tested for and it worked fine.
Java8 worker with Java 11 function app
Function App has references to libraries that are built Java 11 only.

Last part about the dependencies as we are using the same bytecode it will have the same behavior. 
Using these jars for java8 or java11 does not change its behavior. 
Example Netty has no mandatory external dependencies. JDK 1.5 (1.6 for Netty 4+) or above is all that you need to run Netty.

Thank you so much for bringing this point as @trask and I discussed last week a proposed design I am going to test, this proposal should force the customer jars to take presence over worker jars during running customer code, we are working on it.

Please if you have any specific test or concern just let us know.

src/main/java/com/microsoft/azure/functions/worker/broker/FactoryJavaMethodExecutor.java
amamounelsayed: @trask thank you so much for your help. Sounds awesome I will take a note to be updated in the next PR.

pragnagopa:
@amamounelsayed  - Than you for the details. Regarding

> Last part about the dependencies as we are using the same bytecode it will have the same behavior.
Using these jars for java8 or java11 does not change its behavior.
Example Netty has no mandatory external dependencies. JDK 1.5 (1.6 for Netty 4+) or above is all that you need to run Netty.

I understand Java worker 8 will itself load and work as expected in JVM 11. But my concern is specially for the following scenario (using gRpc library as an example) :
- Java Worker 8 has dependency on gRpc package version x which is built for  Java 8
- User wants to use a later version of same gRpc package but version y which might have new API introduced. Again this is an example.
- User runs function app. At runtime, since package referenced by Java Worker takes precedence there is a potential to see "MissingMethodException"

I think until you have a fix for "dependency" problem, this would be the case. If you believe above scenario is not true, then I am ok with the change.  Thanks!

amamounelsayed:
> I understand Java worker 8 will itself load and work as expected in JVM 11. But my concern is specially for the following scenario (using gRpc library as an example) :
> 
> * Java Worker 8 has dependency on gRpc package version x which is built for  Java 8
> * User wants to use a later version of same gRpc package but version y which might have new API introduced. Again this is an example.
> * User runs function app. At runtime, since package referenced by Java Worker takes precedence there is a potential to see "MissingMethodException"
> 
> I think until you have a fix for "dependency" problem, this would be the case. If you believe above scenario is not true, then I am ok with the change. Thanks!

Your scenario is still the case, but this is the same as Java8. So with java11 will currently have the same behavior. This issue will not be fixed with creating new worker with java11 or even new branch. This issue will need to be fixed by changing the classloader behavior which we are working on it for java11 as a new feature to give customers more flexibility, still we are testing approaches to it. 

If the customer uses new library just java11 and we do not have it in worker this should work as it is expected.

amamounelsayed:
@TsuyoshiUshio Thank you for your review, yes we will update it once we are close to roll it out. For the APPVEYOR I totally agree with you, I will work on it for the next PR.

Issue #319, opened 2019-07-31
Support "RpcHttpBodyOnly"capability


yojagad:
Host has a "RpcHttpBodyOnly" capability where it does not send duplicated data through gRPC. It only sends the "body" property which is either bytes or string.

Reference PR: Azure/azure-functions-host#4660

Java needs to make modifications on the worker side accordingly in order to decode it whenever necessary.

Also need to support RpcHttpTriggerMetadataRemoved
Reference https://github.com/Azure/azure-functions-nodejs-worker/pull/238/files

pragnagopa:
cc @mathewc 
@amamounelsayed / @anirudhgarg  - Is it possible to address this issue this sprint? this fix is critical to the perf improvement that @matthewc is working on and specifically he has chosen java worker for making out-of-proc improvements

MatthewC:
@pragnagopa you @ the wrong MatthewC.

anirudhgarg:
Thanks @pragnagopa for flagging this. @TsuyoshiUshio will be helping on this with @amamounelsayed's help.

amamounelsayed:
@pragnagopa @anirudhgarg I talked to @mathewc this fix we do our best to have it in this sprint or next one. Thank you so much!

TsuyoshiUshio:
Hi @yojagad , 
Let me understand the context for the RpcHttpTriggerMetadataRemoved. I read the code on this https://github.com/Azure/azure-functions-nodejs-worker/pull/238/files Inside of the PR, I found the code set the req and $request in triggerMetadata in case of undefined/null.  Why do we need it? I can implement it on Java, however, I want to understand the context.

TsuyoshiUshio:
@yojagad  Self Resolved. :) Never mind. I'll create a PR.

amamounelsayed:
https://github.com/Azure/azure-functions-java-worker/pull/364

TsuyoshiUshio:
@pragnagopa @amamounelsayed Can we close this issue? or keep until the functions host is released?

amamounelsayed:
@TsuyoshiUshio, The reason is not closed it is not released yet. 
@anirudhgarg point to not create confusion, so we close it after release.

MadhuraBharadwaj-MSFT:
Hi @amamounelsayed - could you confirm if this issue can be closed?

PR #364, opened 2020-04-30
Support RpcHttpBodyOnly and RpcHttpTriggerMetadataRemoved

TsuyoshiUshio:
This is the implementation of the issue Support "RpcHttpBodyOnly"capability #319
https://github.com/Azure/azure-functions-java-worker/issues/319

This PR includes: 
* Support RpcHttpBodyOnly capability
* Support RpcHttpTriggerMetadataRemoved
* Mocking gRPC using jMockit
* Resolve the test failure of binding port for passing all test

This is the joint effort with @amamounelsayed 

### RpcHttpTriggerMetadataRemoved
If you don't have RpcHttpTriggerMetadataRemoved, you will find the duplicate data between inputData.0.data_ and triggerMetadata.mapData.$request and "req" These are the identical and requires to send the same json three times. 

image

Then if you specify RpcHttpTriggerMetadataRemoved (and RpcHttpBodyOnly) you will find the triggerMetadata duplication is removed.

image

However, for keeping the same experience for the Azure Functions invocation, we need to copy the "req" and "$request" from inputData.0.data

That is why I implement the method.  This is the same implementation as node. 

## NOTE
The test package structure is not identical as the main package. I recommend to make it the same structure. Then we can use package private for testing methods. Currently I had to make it public just for testability of unit testing. (If you don't like it, I'd like to make it separate class and test it).

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
@@ -3,12 +3,11 @@ package com.microsoft.azure.functions.worker.broker;
 import java.io.File;
 import java.io.IOException;
 import java.net.URL;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
 import com.microsoft.azure.functions.rpc.messages.*;
+import com.microsoft.azure.functions.worker.Constants;
 import com.microsoft.azure.functions.worker.binding.BindingDataStore;
 import com.microsoft.azure.functions.worker.description.FunctionMethodDescriptor;
 import com.microsoft.azure.functions.worker.reflect.ClassLoaderProvider;
@@ -46,7 +45,7 @@ public class JavaFunctionBroker {
 
 		BindingDataStore dataStore = new BindingDataStore();
 		dataStore.setBindingDefinitions(executor.getBindingDefinitions());
-		dataStore.addTriggerMetadataSource(request.getTriggerMetadataMap());
+		dataStore.addTriggerMetadataSource(getTriggerMetadataMap(request));
 		dataStore.addParameterSources(request.getInputDataList());
 		dataStore.addExecutionContextSource(request.getInvocationId(), methodEntry.left, request.getTraceContext().getTraceParent(), request.getTraceContext().getTraceState(), request.getTraceContext().getAttributesMap());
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
```diff
@@ -21,7 +21,8 @@
 	public RpcHttpRequestDataSource(String name, RpcHttp value) {
 		super(name, null, HTTP_DATA_OPERATIONS);
 		this.httpPayload = value;
-		this.bodyDataSource = BindingDataStore.rpcSourceFromTypedData(null, this.httpPayload.getBody());
+//		this.bodyDataSource = BindingDataStore.rpcSourceFromTypedData(null, this.httpPayload.getBody());
```
pragnagopa, line 24: nit: remove commented code.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -59,6 +57,29 @@ public void loadMethod(FunctionMethodDescriptor descriptor, Map<String, BindingI
...
+				name = e.getName();
+				break;
+			}
+		}
+
+		Map<String, TypedData> triggerMetadata = new HashMap(request.getTriggerMetadataMap());
+		if (!name.isEmpty() && !triggerMetadata.containsKey(name)) {
+			triggerMetadata.put(name, dataWithHttp);
+		}
+		String requestKey = "$request";
```
pragnagopa, line 76: nit: move to constants file

mathewc:
In my PR https://github.com/Azure/azure-functions-host/pull/5897, I'm no longer sending ANY duplicate htttp request data via BindingData, since the request details are already sent as part of inputs. With your changes, will things work with none of these request details being in the binding data?

TsuyoshiUshio:
@mathewc Yes. It works. My code takes value from inputData to create triggerMetadate like req, $request.  I need research how Java Azure Functions side uses these triggerMetadata, however, if we can create the triggerMetadata from inputData, it must be fine. I'll do the research for the usage of the triggerMetadata and will share in here.

src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
TsuyoshiUshio: Done

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
TsuyoshiUshio: Done

TsuyoshiUshio:
@pragnagopa I've done the fix. Also I add one more test scenario (if the request is non-http request, no trigger metadata is modified). @amamounelsayed if you need to discuss more, please leave your message. If @amamounelsayed  has no comment, let's get this pr merged!

amamounelsayed:
Thank you @TsuyoshiUshio I will review it and merge sir.

amamounelsayed:
@TsuyoshiUshio I found an issue with RpcHttpBodyOnly  logic, that is initially I added. I fixed it. I am writing tests for it.

PR #379, opened 2020-06-17
Feature/shadding

### LINKED ISSUE #183: Investigate dependecy conflicts between worker and user code

### LINKED ISSUE #340: Worker dependencies must be shaded (package renamed) to avoid dependency collision
Looking at this https://github.com/Azure/azure-functions-java-worker/wiki/Azure-Java-Functions-Worker-Dependencies, couldn't you shade common libraries in order to avoid dependency collision and runtime issue ?

I am talking about guava, netty, apache commons, ...

Right now, it is very difficult to understand why a function is failing when it happens.

I tried to use azure storage java sdk v 12.1.0 (last version) and because of a collision with netty it just could not work (Luckily the sdk provides a way to also use the okhttp client else I would have been stuck)

### LINKED ISSUE #341: Function Fails on NoClassDefFoundError
#### Investigative information

Please provide the following:

- Timestamp: 2020-01-07 14:25:39.9578344
- Function App name: please contact me for more information
- Function name(s) (as appropriate): processUPSNucasCarrierUpdateFunction
- Invocation ID:
- Region: Central US (waws-prod-dm1-133)

#### Expected behavior

The function app suppose to work and continue handling EventHub requests.

#### Actual behavior

This issue is transient, the function stops working after initialization.
.Jar modules import completes before handling request.
First EventHub request fails on the Java function instance.
```java
Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException : Result: Failure
Exception: NoClassDefFoundError: Could not initialize class xxx.xxx.xxx.datasource.statusmappingcache.CtaUpsCarrierStatusMappingCache
Stack: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
```

#### Known workarounds

None. The issue is transient.

#### Related information 

I check the storage account, and connection to it, everything is fine.
Function package is deployed using WEBSITE_RUN_FROM_PACKAGE=1, all content mounting operations happen on the VM memory, network is not a concern.

The .jar files does contain the classes, it could be a **module import** on our Java worker, or a generic Zulu 8.36.0.1-jre8.0.202 JVM issue.

Please contact @Hazhzeng for more information on customer's app.

cc: @ankitkumarr @pragnagopa 
</details>
-->

### LINKED ISSUE #365: Java 3rd Party dependency conflict between Azure Function and Azure SDK for Java
# Technical Details

When using **Azure Function** and [**Azure SDK for Java**](https://github.com/Azure/azure-sdk-for-java) simultaneously. It causes a [**dependency conflict**](https://github.com/Azure/azure-sdk-for-java/issues/8392).
For example, using Azure Function with App Configuration SDK for Java as dependencies in an independent app, 
**Expected behavior**: running the application successfully. 
**Actual behavior**: application crashed with unhandled exception, 
```
[2/21/2020 9:42:14 PM] Executed 'Functions.hello' (Failed, Id=24c8735c-df65-45ea-a857-8ac83005b7fa)
[2/21/2020 9:42:14 PM] System.Private.CoreLib: Exception while executing function: Functions.hello. System.Private.CoreLib: Result: Failure
Exception: NoSuchMethodError: io.netty.handler.ssl.SslProvider.isAlpnSupported(Lio/netty/handler/ssl/SslProvider;)Z
Stack: java.lang.reflect.InvocationTargetException
[2/21/2020 9:42:14 PM]  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[2/21/2020 9:42:14 PM]  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[2/21/2020 9:42:14 PM]  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[2/21/2020 9:42:14 PM]  at java.lang.reflect.Method.invoke(Method.java:498)
[2/21/2020 9:42:14 PM]  at com.microsoft.azure.functions.worker.broker.JavaMethodInvokeInfo.invoke(JavaMethodInvokeInfo.java:22)
[2/21/2020 9:42:14 PM]  at com.microsoft.azure.functions.worker.broker.JavaMethodExecutor.execute(JavaMethodExecutor.java:54)
[2/21/2020 9:42:14 PM]  at com.microsoft.azure.functions.worker.broker.JavaFunctionBroker.invokeMethod(JavaFunctionBroker.java:53)
[2/21/2020 9:42:14 PM]  at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:33)
[2/21/2020 9:42:14 PM]  at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:10)
[2/21/2020 9:42:14 PM]  at com.microsoft.azure.functions.worker.handler.MessageHandler.handle(MessageHandler.java:45)
[2/21/2020 9:42:14 PM]  at com.microsoft.azure.functions.worker.JavaWorkerClient$StreamingMessagePeer.lambda$onNext$0(JavaWorkerClient.java:92)
[2/21/2020 9:42:14 PM]  at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[2/21/2020 9:42:14 PM]  at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[2/21/2020 9:42:14 PM]  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[2/21/2020 9:42:14 PM]  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[2/21/2020 9:42:14 PM]  at java.lang.Thread.run(Thread.java:748)
[2/21/2020 9:42:14 PM] Caused by: java.lang.NoSuchMethodError: io.netty.handler.ssl.SslProvider.isAlpnSupported(Lio/netty/handler/ssl/SslProvider;)Z
[2/21/2020 9:42:14 PM]  at reactor.netty.http.client.HttpClientSecure.<clinit>(HttpClientSecure.java:79)
```
More details of the exception can redirect to the Azure SDK for Java issue: https://github.com/Azure/azure-sdk-for-java/issues/8392
#  Investigation
`Azure Function` is using [`GRPC-netty 1.20.0`](https://mvnrepository.com/artifact/io.grpc/grpc-netty/1.20.0) which using `Netty 4.1.34.Final`. The latest version of GRPC-Netty is [`1.28.0`](https://mvnrepository.com/artifact/io.grpc/grpc-netty/1.28.0) which using version `4.1.45.Final` for both artifacts, `netty-codec-http2` and `netty-handler-proxy` in the io.netty group.

`Azure SDK for Java` is using `Netty 4.1.44.Final`. (I am from Azure SDK for Java team, we are going to upgrade to `Netty 4.1.45.Final` for above two artifacts, `netty-codec-http2` and `netty-handler-proxy` in the 2020 April release, which is in the first week of April. See section `Dependencies` for more details on other artifacts of io.netty and version that we are going to use.). 

#  Proposal
Upgrade [`GRPC-netty 1.20.0`](https://mvnrepository.com/artifact/io.grpc/grpc-netty/1.20.0) to latest version [`1.28.0`](https://mvnrepository.com/artifact/io.grpc/grpc-netty/1.28.0) will make `Azure Function` and `Azure SDK for Java` dependency consistency because  both SDKs are using the same Netty version `4.1.45` then. 

# Dependencies
**AZURE FUNCTION** dependency tree:
```
com.microsoft.azure.functions:azure-functions-java-worker:jar:1.5.2
[INFO] +- com.microsoft.azure.functions:azure-functions-java-library:jar:1.3.1-SNAPSHOT:compile
[INFO] +- com.google.protobuf:protobuf-java:jar:3.7.1:compile
[INFO] +- io.grpc:grpc-protobuf:jar:1.20.0:compile
[INFO] |  +- io.grpc:grpc-core:jar:1.20.0:compile
[INFO] |  |  +- io.grpc:grpc-context:jar:1.20.0:compile
[INFO] |  |  +- com.google.code.gson:gson:jar:2.8.5:compile
[INFO] |  |  +- com.google.errorprone:error_prone_annotations:jar:2.3.2:compile
[INFO] |  |  +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO] |  |  +- com.google.android:annotations:jar:4.1.1.4:compile
[INFO] |  |  +- org.codehaus.mojo:animal-sniffer-annotations:jar:1.17:compile
[INFO] |  |  +- io.opencensus:opencensus-api:jar:0.19.2:compile
[INFO] |  |  \- io.opencensus:opencensus-contrib-grpc-metrics:jar:0.19.2:compile
[INFO] |  +- com.google.guava:guava:jar:26.0-jre:compile
[INFO] |  |  +- org.checkerframework:checker-qual:jar:2.5.2:compile
[INFO] |  |  \- com.google.j2objc:j2objc-annotations:jar:1.1:compile
[INFO] |  +- com.google.api.grpc:proto-google-common-protos:jar:1.12.0:compile
[INFO] |  \- io.grpc:grpc-protobuf-lite:jar:1.20.0:compile
[INFO] +- io.grpc:grpc-stub:jar:1.20.0:compile
[INFO] +- io.grpc:grpc-netty:jar:1.20.0:compile
[INFO] |  +- io.netty:netty-codec-http2:jar:4.1.34.Final:compile (version selected from constraint [4.1.34.Final,4.1.34.Final])
[INFO] |  |  +- io.netty:netty-common:jar:4.1.34.Final:compile
[INFO] |  |  +- io.netty:netty-buffer:jar:4.1.34.Final:compile
[INFO] |  |  +- io.netty:netty-transport:jar:4.1.30.Final:compile
[INFO] |  |  |  \- io.netty:netty-resolver:jar:4.1.30.Final:compile
[INFO] |  |  +- io.netty:netty-codec:jar:4.1.34.Final:compile
[INFO] |  |  +- io.netty:netty-handler:jar:4.1.30.Final:compile
[INFO] |  |  \- io.netty:netty-codec-http:jar:4.1.30.Final:compile
[INFO] |  \- io.netty:netty-handler-proxy:jar:4.1.30.Final:compile
[INFO] |     \- io.netty:netty-codec-socks:jar:4.1.30.Final:compile
[INFO] +- org.apache.commons:commons-lang3:jar:3.9:compile
[INFO] +- commons-cli:commons-cli:jar:1.4:compile
[INFO] +- javax.annotation:javax.annotation-api:jar:1.3.2:compile
[INFO] +- junit:junit:jar:4.12:test
[INFO] |  \- org.hamcrest:hamcrest-core:jar:1.3:test
[INFO] \- net.java.dev.jna:jna-platform:jar:5.3.0:compile
[INFO]    \- net.java.dev.jna:jna:jar:5.3.0:compile
```
**Azure SDK for Java** dependency tree and the maven **io.netty dependencies** will be used in the upcoming April release.:
```
// Current Azure SDK's io.netty dependency version
 com.azure:azure-core-http-netty:jar:1.4.0:compile
[INFO] |  +- io.netty:netty-handler:jar:4.1.44.Final:compile
[INFO] |  |  +- io.netty:netty-common:jar:4.1.44.Final:compile
[INFO] |  |  +- io.netty:netty-transport:jar:4.1.44.Final:compile
[INFO] |  |  |  \- io.netty:netty-resolver:jar:4.1.44.Final:compile
[INFO] |  |  \- io.netty:netty-codec:jar:4.1.44.Final:compile
[INFO] |  +- io.netty:netty-handler-proxy:jar:4.1.44.Final:compile
[INFO] |  |  \- io.netty:netty-codec-socks:jar:4.1.44.Final:compile
[INFO] |  +- io.netty:netty-buffer:jar:4.1.44.Final:compile
[INFO] |  +- io.netty:netty-codec-http:jar:4.1.44.Final:compile
[INFO] |  +- io.netty:netty-codec-http2:jar:4.1.44.Final:compile
[INFO] |  +- io.netty:netty-transport-native-unix-common:jar:4.1.44.Final:compile
[INFO] |  +- io.netty:netty-transport-native-epoll:jar:linux-x86_64:4.1.44.Final:compile
[INFO] |  \- io.projectreactor.netty:reactor-netty:jar:0.9.4.RELEASE:compile

// Upcoming Azure SDK's io.netty dependency version
<dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-handler</artifactId>
      <version>4.1.45.Final</version> 
    </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-handler-proxy</artifactId>
      <version>4.1.45.Final</version> 
    </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-buffer</artifactId>
      <version>4.1.45.Final</version> 
    </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-codec-http</artifactId>
      <version>4.1.45.Final</version> 
    </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-codec-http2</artifactId>
      <version>4.1.45.Final</version> 
    </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-transport-native-unix-common</artifactId>
      <version>4.1.45.Final</version> 
    </dependency>
    <dependency>
      <groupId>io.netty</groupId>
      <artifactId>netty-transport-native-epoll</artifactId>
      <version>4.1.45.Final</version>
      <classifier>linux-x86_64</classifier>
    </dependency>
```

cc: @mrm9084

### LINKED ISSUE #368: WebClient of Spring Boot (3rd dependency) conflicts between Azure Function and Azure SDK for Java
**Describe the bug**
When calling webClient in an azure function runtime locally,  the log will print an error.

#### Investigative information
**Exception or Stack Trace**
`[05/11/2020 15:24:58] System.Private.CoreLib: Exception while executing function: Functions.RunMyHttpTrigger. System.Private.CoreLib: Result: Failure
[05/11/2020 15:24:58] Exception: NoSuchMethodError: io.netty.handler.ssl.SslProvider.isAlpnSupported(Lio/netty/handler/ssl/SslProvider;)Z
[05/11/2020 15:24:58] Stack: java.lang.reflect.InvocationTargetException
[05/11/2020 15:24:58]   at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[05/11/2020 15:24:58]   at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[05/11/2020 15:24:58]   at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[05/11/2020 15:24:58]   at java.lang.reflect.Method.invoke(Method.java:498)
[05/11/2020 15:24:58]   at com.microsoft.azure.functions.worker.broker.JavaMethodInvokeInfo.invoke(JavaMethodInvokeInfo.java:22)
[05/11/2020 15:24:58]   at com.microsoft.azure.functions.worker.broker.JavaMethodExecutor.execute(JavaMethodExecutor.java:54)
[05/11/2020 15:24:58]   at com.microsoft.azure.functions.worker.broker.JavaFunctionBroker.invokeMethod(JavaFunctionBroker.java:53)
[05/11/2020 15:24:58]   at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:33)
[05/11/2020 15:24:58]   at com.microsoft.azure.functions.worker.handler.InvocationRequestHandler.execute(InvocationRequestHandler.java:10)
[05/11/2020 15:24:58]   at com.microsoft.azure.functions.worker.handler.MessageHandler.handle(MessageHandler.java:45)
[05/11/2020 15:24:58]   at com.microsoft.azure.functions.worker.JavaWorkerClient$StreamingMessagePeer.lambda$onNext$0(JavaWorkerClient.java:92)
[05/11/2020 15:24:58]   at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[05/11/2020 15:24:58]   at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[05/11/2020 15:24:58]   at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[05/11/2020 15:24:58]   at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[05/11/2020 15:24:58]   at java.lang.Thread.run(Thread.java:748)
[05/11/2020 15:24:58] Caused by: java.lang.NoSuchMethodError: io.netty.handler.ssl.SslProvider.isAlpnSupported(Lio/netty/handler/ssl/SslProvider;)Z
[05/11/2020 15:24:58]   at reactor.netty.http.client.HttpClientSecure.<clinit>(HttpClientSecure.java:79)
[05/11/2020 15:24:58]   at reactor.netty.http.client.HttpClientConnect$MonoHttpConnect.lambda$subscribe$0(HttpClientConnect.java:298)
[05/11/2020 15:24:58]   at reactor.core.publisher.MonoCreate.subscribe(MonoCreate.java:57)
[05/11/2020 15:24:58]   at reactor.core.publisher.FluxRetryPredicate$RetryPredicateSubscriber.resubscribe(FluxRetryPredicate.java:124)
[05/11/2020 15:24:58]   at reactor.core.publisher.MonoRetryPredicate.subscribeOrReturn(MonoRetryPredicate.java:51)
[05/11/2020 15:24:58]   at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:48)
[05/11/2020 15:24:58]   at reactor.netty.http.client.HttpClientConnect$MonoHttpConnect.subscribe(HttpClientConnect.java:323)
[05/11/2020 15:24:58]   at reactor.core.publisher.InternalMonoOperator.subscribe(InternalMonoOperator.java:55)
[05/11/2020 15:24:58]   at reactor.core.publisher.MonoDefer.subscribe(MonoDefer.java:52)
[05/11/2020 15:24:58]   at reactor.core.publisher.Mono.subscribe(Mono.java:4210)
[05/11/2020 15:24:58]   at reactor.core.publisher.Mono.block(Mono.java:1665)
[05/11/2020 15:24:58]   at de.hybris.azure.trigger.eventtrigger.runMyHttpTrigger.obtainSecuredResource(RunMyHttpTrigger.java:1234)
[05/11/2020 15:24:58]   at de.hybris.azure.trigger.eventtrigger.runMyHttpTrigger.run(RunMyHttpTrigger.java:1230)
[05/11/2020 15:24:58]   ... 16 more
[05/11/2020 15:24:58] .
`
 
**Reproduce the issue**
Build a function (see code below)
Run mvn clean package azure-functions:run
Call the generated api. The api call will timeout and a stacktrace will be printed.

**Code snippet**
`public class RunMyHttpTrigger extends AzureSpringBootRequestHandler<HttpRequestMessage, CloseableHttpResponse>
{
    @FunctionName("runMyHttpTrigger")
    public HttpResponseMessage run(
            @HttpTrigger(name = "req", methods = {HttpMethod.GET, HttpMethod.POST}, authLevel = AuthorizationLevel.ANONYMOUS) HttpRequestMessage<Optional<String>> request,
            final ExecutionContext context) throws IOException
    {
        context.getLogger().info("Java HTTP trigger processed a request.");

        obtainSecuredResource(context);
        return  request.createResponseBuilder(HttpStatus.BAD_REQUEST).header("Content-Type","application/json").body("Bad Request.").build();
         
    }

    public Mono<String> obtainSecuredResource(final ExecutionContext context)
    {
        final String tokenURI = "https://abc.def.com/oauth/token";
        final String clientId = "clientId";
        final String clientSecret = "clientSecret";
        final String endPointURI = "https://abc.def.com/registration";

        context.getLogger().info("=========================================\n");
        context.getLogger().info("Access web client...");
        Mono<String> resource = WebClient.builder()
                .build().post().uri(tokenURI).header(HttpHeaders.AUTHORIZATION, "Basic " + Base64Utils
                        .encodeToString((clientId + ":" + clientSecret).getBytes())).body(
                        BodyInserters.fromFormData(OAuth2ParameterNames.GRANT_TYPE, GrantType.CLIENT_CREDENTIALS.getValue())).retrieve()
                .onStatus(org.springframework.http.HttpStatus::is4xxClientError, response -> Mono.empty()).bodyToMono(String.class).doOnNext(
                        System.out::println);
       //exception happened this line:
        context.getLogger().info("non-blocking Oauth calls registered..." + resource.block());
        //resource.block();
        //return "test";
        return resource.map(res -> "Retrieved the resource using a manual approach: " + res);
    }
}`

In pom.xml, I have explicit the dependency about all io.netty with version 4.1.49.Final, and io.grpc with version 1.29.0.

Setup (please complete the following information):

OS: MacOS
IDE :InteliJ/Terminal

### LINKED ISSUE #373: Epoll based Netty Transport error in Function Development Tool on Linux
Function Development Tool on Linux (I guess this is true for production Linux Function as well) fails when using "Epoll based Netty Transport" in the application. The issue can be reproduced by creating a function app with netty modules as it dependency along with "netty-Epoll-for-Linux".

### Steps to reproduce:

Create an Azure Ubuntu 18.04 Linux VM.
##### Set Package feed:
```
wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb
sudo dpkg -i packages-microsoft-prod.deb
sudo apt-get update
   ```

##### Install Function tool:

```
sudo apt-get install azure-functions-core-tools-3
```

##### Intstall Java8:

```
sudo add-apt-repository --remove ppa:webupd8team/java
sudo apt install apt-transport-https && sudo apt update && sudo apt install -y openjdk-8-jdk
```

###### Find path to Java 8:
```
sudo update-alternatives --config java
Example: > /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
```

###### Open environment file:

```
sudo vi /etc/environment
```

###### Set Java home, save and exit:

```
JAVA_HOME="/usr/lib/jvm/java-8-openjdk-amd64"
```
###### Reload environment: 

```
source /etc/environment
```

##### Install Maven:

```
sudo apt install maven
```

#### Download repro-function project to this Linux box, package and Run:

[azure-func-netty.zip](https://github.com/Azure/azure-functions-java-worker/files/4664279/azure-func-netty.zip)

```
mvn clean package
mvn azure-functions:run &
```

#### Hit function endpoint:

```
curl http://localhost:7071/api/HttpExample?name=hello
```

You should get the below error:

```
Caused by: java.lang.NoSuchMethodError: 
io.netty.channel.SingleThreadEventLoop.<init>(Lio/netty/channel/EventLoopGroup;Ljava/util/concurrent/Executor;ZLjava/util/Queue;Ljava/util/Queue;Lio/netty/util/concurrent/RejectedExecutionHandler;)V
	at io.netty.channel.epoll.EpollEventLoop.<init>(EpollEventLoop.java:87)
	at io.netty.channel.epoll.EpollEventLoopGroup.newChild(EpollEventLoopGroup.java:143)
	at io.netty.channel.epoll.EpollEventLoopGroup.newChild(EpollEventLoopGroup.java:35)
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:84)
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:58)
	at io.netty.util.concurrent.MultithreadEventExecutorGroup.<init>(MultithreadEventExecutorGroup.java:47)
	at io.netty.channel.MultithreadEventLoopGroup.<init>(MultithreadEventLoopGroup.java:59)
	at io.netty.channel.epoll.EpollEventLoopGroup.<init>(EpollEventLoopGroup.java:104)
	at io.netty.channel.epoll.EpollEventLoopGroup.<init>(EpollEventLoopGroup.java:91)
	at io.netty.channel.epoll.EpollEventLoopGroup.<init>(EpollEventLoopGroup.java:68)
	at io.netty.channel.epoll.EpollEventLoopGroup.<init>(EpollEventLoopGroup.java:52)
	at io.netty.channel.epoll.EpollEventLoopGroup.<init>(EpollEventLoopGroup.java:45)
	at org.example.Function.doHttpGet(Function.java:71)
	at org.example.Function.run(Function.java:55)
	... 16 more
```

#### Here is a normal console app, having the same netty dependencies but runs fine

[linux-use-netty1.zip](https://github.com/Azure/azure-functions-java-worker/files/4664294/linux-use-netty1.zip)

Download this project to the above Linux box and run:

```
mvn clean compile exec:java
```

The Epoll based code work as expected.

### PR DESCRIPTION
**Pull request by amamounelsayed**
This feature is for shading all worker jars.

Only for JAVA 8
To be a backward compatible I freeze all the jars we currently have added to repo.
Then when we load the classloader we will load the worker lib first then customer.

If the user like to reverse this order and their jars takes presence first, they will need to add FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS True or 1. By this way it will be opt-in scenario.

JAVA 11 will use only the customer jars.

Generated jar after shading:
[azure-functions-java-worker.zip](https://github.com/Azure/azure-functions-java-worker/files/4800176/azure-functions-java-worker.zip)

Fixes #340, Fixes #183, Fixes #373, Fixes #368, Fixes #365, Fixes #341

### CODE REVIEW

File: src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java [hunk 1]
```diff
@@ -89,7 +93,45 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
 
 	private void registerWithClassLoaderProvider(File libDirectory) {
 		try {
-			classLoaderProvider.addDirectory(libDirectory);
+			if(SystemUtils.IS_JAVA_1_8) {
+				String workerPath = System.getenv(Constants.FUNCTIONS_WORKER_DIRECTORY) + "/lib";
+				String reverseJarsLoading = System.getenv(Constants.JAVA_REVERSE_JARS_LOADING);
+				if(reverseJarsLoading != null && reverseJarsLoading.toLowerCase().equals("true")) {
+					if(!new File(workerPath).exists()) {
```
amamounelsayed, line 100: Should we dramatically throw exception if we do not have access to worker lib? 
My opinion is no. But need to hear feedback.

amamounelsayed: I talked to @fabiocav and he suggested to have it sever.

src/main/java/com/microsoft/azure/functions/worker/Constants.java
```diff
@@ -6,4 +6,6 @@
 public final class Constants {
     private Constants(){}
     public final static String TRIGGER_METADATA_DOLLAR_REQUEST_KEY = "$request";
+    public final static String FUNCTIONS_WORKER_DIRECTORY = "FUNCTIONS_WORKER_DIRECTORY";
+    public final static String JAVA_REVERSE_LIB_LOADING = "JAVA_REVERSE_LIB_LOADING";
```
pragnagopa, line 10: Rename to : `FUNCTIONS_WORKER_JAVA_USE_SHADED_LIBRARIES` - to make this more explicit and is a java worker behavior

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +91,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
+			throw new Exception("Error loading worker jars, for "+ status +", from path:  " + workerLibPath);
+		} else {
+			File[] jarFiles = workerLib.listFiles(new FileFilter() {
+				@Override
+				public boolean accept(File file) {
+					return file.isFile() && file.getName().endsWith(".jar");
+				}
+			});
+			if(jarFiles.length == 0) {
+				throw new Exception("Error loading worker jars, for "+ status +", from path:  " + workerLibPath + ". Jars size is zero");
```
pragnagopa, line 130: should this throw? or just return false?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +91,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
+				classLoaderProvider.addDirectory(libDirectory);
+			}
 		} catch (Exception ex) {
 			ExceptionUtils.rethrow(ex);
 		}
 	}
 
+	boolean checkLibFolder(File workerLib, String workerLibPath, String status) throws Exception{
+		if(!workerLib.exists()) {
+			throw new Exception("Error loading worker jars, for "+ status +", from path:  " + workerLibPath);
```
pragnagopa, line 121: Is this always guaranteed to exist? Also, instead of generic exception,  throw FileNotFoundException?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +91,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
+			} else {
+				classLoaderProvider.addDirectory(libDirectory);
+			}
 		} catch (Exception ex) {
 			ExceptionUtils.rethrow(ex);
 		}
 	}
 
+	boolean checkLibFolder(File workerLib, String workerLibPath, String status) throws Exception{
+		if(!workerLib.exists()) {
```
pragnagopa, line 120: `status` is only used for logging, remove this is input parameter and handle logging in the if..else above.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +91,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
+				}
+			} else {
+				classLoaderProvider.addDirectory(libDirectory);
+			}
 		} catch (Exception ex) {
 			ExceptionUtils.rethrow(ex);
 		}
 	}
 
+	boolean checkLibFolder(File workerLib, String workerLibPath, String status) throws Exception{
```
pragnagopa, line 119: rename to throwIfFolderNotExisits?

src/main/java/com/microsoft/azure/functions/worker/Constants.java
amamounelsayed: We will use shaded jars regardless, this setting to load which jars first, worker or client.
What about FUNCTIONS_WORKER_JAVA_CLIENT_LIB_FIRST?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
amamounelsayed: This should throw exception to guarantee that the worker lib folder contains jars not just an empty folder.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
amamounelsayed: You have a point the reason is as the exception is created in the method, this method now serve two purposes, exception and exception message. To change the message there is three ways, the one proposed.
second return just false and pass the excptionString function parameter and create the exception in the else.What do you think?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
amamounelsayed: Will change it to throwJarsNotExists as it is serving if you have the folder but empty.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
amamounelsayed: Will change it to FileNotFoundException. 
You bring good point, What is the other cases that can cause it not to exit?

endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/HttpTriggerTests.java
```diff
@@ -108,4 +103,31 @@ public HttpResponseMessage HttpTriggerCustomCode(
...
+            final ExecutionContext context
+    ) throws Exception {
+        context.getLogger().info("Java HTTP trigger processed a request.");
+
+        // Parse query parameters
+        String query = request.getQueryParameters().get("name");
+        String name = request.getBody().orElse(query);
+        String readEnv = System.getenv("AzureWebJobsStorage");
+
+        Gson a = new Gson();
```
TsuyoshiUshio, line 119: What is the purpose of this code? "a" is not referred from anywhere.

src/main/java/com/microsoft/azure/functions/worker/Constants.java
pragnagopa: `FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS`  -- this means that libs from app take precedence. No need to add First?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +92,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
 
-	private void registerWithClassLoaderProvider(File libDirectory) {
+	void registerWithClassLoaderProvider(File libDirectory) {
 		try {
-			classLoaderProvider.addDirectory(libDirectory);
+			if(SystemUtils.IS_JAVA_1_8) {
+				String workerLibPath = System.getenv(Constants.FUNCTIONS_WORKER_DIRECTORY) + "/lib";
+				String javaReverseLibLoading = System.getenv(Constants.JAVA_REVERSE_LIB_LOADING);
+				File workerLib = new File(workerLibPath);
+				if(javaReverseLibLoading != null && javaReverseLibLoading.toLowerCase().equals("true")) {
```
pragnagopa, line 101: Move this to a helper.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +92,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
+				}
+			} else {
+				classLoaderProvider.addDirectory(libDirectory);
+			}
 		} catch (Exception ex) {
 			ExceptionUtils.rethrow(ex);
 		}
 	}
 
+	boolean throwJarsNotFound(File workerLib, String workerLibPath, String status) throws FileNotFoundException{
```
pragnagopa, line 120: this method does not have to return bool. This is only checking if jars exist or not. Change to `verifyLibrariesExist` ?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
pragnagopa: This is mostly a Nit. Up to you.

src/main/java/com/microsoft/azure/functions/worker/Constants.java
amamounelsayed: Sounds good, I will change it, the reason I added first that in both cases we load the jars one take presence over the other.

endtoendtests/Azure.Functions.Java.Tests.E2E/Azure.Functions.Java.Tests.E2E/HttpEndToEndTests.cs
```diff
@@ -33,5 +34,20 @@ public async Task HttpTrigger_ReturnsCustomCode()
         {
             Assert.True(await Utilities.InvokeHttpTrigger("HttpTriggerCustomCode", "?&name=Test",  HttpStatusCode.OK, "Test", 209));
         }
+
+        [Fact]
+        public async Task HttpTriggerJavaClassLoader()
```
TsuyoshiUshio, line 39: Which scenario would you want to validate? 
According to the code, this test validates, if the environment variables "AzureWebJobsStrage" is exists or not when the java_home does not contain "zulu-11" and "JAVA_REVERSE_LIB_LOADING" is false or null.  In this case, you are changing the order of the class loading. So the test should be valid if you change the "JAVA_REVERSE_LIB_LOADING"  configuration, the order is changed properly.

endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/HttpTriggerTests.java
```diff
@@ -47,11 +47,6 @@ public HttpResponseMessage HttpTriggerJava(
         }
 
         Gson a = new Gson();
```
TsuyoshiUshio, line 49: Why this instantiation is here?

package.ps1
```diff
@@ -23,6 +23,7 @@ Get-ChildItem -Path .\target\* -Include 'azure*' -Exclude '*shaded.jar' | %{ Cop
 StopOnFailedExecution
 copy-item ./worker.config.json pkg
 copy-item ./tools/AzureFunctionsJavaWorker.nuspec pkg/
+Copy-Item ".\lib_worker_1.6.2" pkg\lib -Recurse
```
TsuyoshiUshio, line 26: According to the `pom.xml` version is 1.7.1, not 1.6.2. If the number `1.6.2` will be changed in the future, we might better consider it in one place. I can see the several 1.6.2 on the pipeline.

pom.xml
```diff
@@ -203,6 +204,56 @@
 									</excludes>
 								</filter>
 							</filters>
+							<relocations>
```
TsuyoshiUshio, line 207: You look sharded everything. Then if we add a new library, you want to shaded as well right? 
Then it might be better to add the direction to the developer's guide.

package.ps1
amamounelsayed: The reason for 1.6.2 this is the jars for the latest release we currently have.

endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/HttpTriggerTests.java
amamounelsayed: I just wanted to make sure that there is no class exception. I will add more comments.

endtoendtests/Azure.Functions.Java.Tests.E2E/Azure.Functions.Java.Tests.E2E/HttpEndToEndTests.cs
amamounelsayed: I will more explanation sir

src/main/java/com/microsoft/azure/functions/worker/Utilities.java
```diff
@@ -0,0 +1,10 @@
+package com.microsoft.azure.functions.worker;
+
+public class Utilities {
+    public static boolean isTrue(String value) {
+        if(value != null && value.toLowerCase().equals("true")) {
```
pragnagopa, line 5: Rename to `isLoadAppLibs` .  Make this method specific to checking if `FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS` is set to true or "1"

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +93,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
-	private void registerWithClassLoaderProvider(File libDirectory) {
+	void registerWithClassLoaderProvider(File libDirectory) {
 		try {
-			classLoaderProvider.addDirectory(libDirectory);
+			if(SystemUtils.IS_JAVA_1_8) {
+				String workerLibPath = System.getenv(Constants.FUNCTIONS_WORKER_DIRECTORY) + "/lib";
+				String javaReverseLibLoading = System.getenv(Constants.FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS);
+				File workerLib = new File(workerLibPath);
+				if(Utilities.isTrue(javaReverseLibLoading)) {
+					if(verifyLibrariesExist (workerLib, workerLibPath, "CLIENT_LIB_FIRST")) {
```
pragnagopa, line 103: it should be enough to call `verifyLibrariesExist` once right? both calls use same arguments?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +93,48 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
+						classLoaderProvider.addDirectory(libDirectory);
+						classLoaderProvider.addDirectory(workerLib);
+					}
+				} else {
+					if(verifyLibrariesExist (workerLib, workerLibPath, "WORKER_LIB_FIRST")) {
```
pragnagopa, line 108: Change to
```java
if(SystemUtils.IS_JAVA_1_8) {
				String workerLibPath = System.getenv(Constants.FUNCTIONS_WORKER_DIRECTORY) + "/lib";
				String javaReverseLibLoading = System.getenv(Constants.FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS);
				File workerLib = new File(workerLibPath);
                verifyLibrariesExist (workerLib, workerLibPath);
				if(Utilities.isTrue(javaReverseLibLoading)) {
						classLoaderProvider.addDirectory(libDirectory);
						classLoaderProvider.addDirectory(workerLib);
					}else {
						classLoaderProvider.addDirectory(workerLib);
						classLoaderProvider.addDirectory(libDirectory);
					}
				}
```

pragnagopa: move this to a different method. Make it much easier to read code. Move `SystemUtils.IS_JAVA_1_8` to utilities as well : `IsJava8`

src/main/java/com/microsoft/azure/functions/worker/Utilities.java
amamounelsayed: I created as Utilities so we can use it anywhere in future. I believe much better to be more generic as the logic is not specific for FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS .

The file should name util my fault.

pragnagopa: having static methods that sepcific to checking environment is useful. Please see here for example pattern: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Environment/EnvironmentExtensions.cs also `isTrue` is very generic. I recommend making specific helpers.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
amamounelsayed: The arguments are different but you raised good point we do not need to have the last param. I will make it once call.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
amamounelsayed: Please have a look on the next code update, I believe it is more clean. I just prefer to be in one place for more easy to follow. And Util class has only generic methods.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +91,49 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
+				}
+			} else {
+				classLoaderProvider.addDirectory(libDirectory);
+			}
 		} catch (Exception ex) {
 			ExceptionUtils.rethrow(ex);
 		}
 	}
 
+	boolean verifyLibrariesExist (File workerLib, String workerLibPath) throws FileNotFoundException{
```
pragnagopa, line 120: this method never returns false. make this void. If calling this method does not throw , the it is ok.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -87,14 +91,49 @@ private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) thro
...
-	private void registerWithClassLoaderProvider(File libDirectory) {
+	void registerWithClassLoaderProvider(File libDirectory) {
 		try {
-			classLoaderProvider.addDirectory(libDirectory);
+			if(SystemUtils.IS_JAVA_1_8) {
+				String workerLibPath = System.getenv(Constants.FUNCTIONS_WORKER_DIRECTORY) + "/lib";
+				String javaReverseLibLoading = System.getenv(Constants.FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS);
+				File workerLib = new File(workerLibPath);
+
+				if(verifyLibrariesExist (workerLib, workerLibPath)) {
```
pragnagopa, line 101: Just call verifyLibrariesExist. this will simplify the code.

src/main/java/com/microsoft/azure/functions/worker/Utilities.java
amamounelsayed: Good point I added a helper for that.

pragnagopa:
One more question: instead of checking in individual jars, is it possible to ship one jar file?

amamounelsayed:
> One more question: instead of checking in individual jars, is it possible to ship one jar file?

I am sorry did not get the question.

pragnagopa:
> I am sorry did not get the question.

This PR includes individual jar files in folder `lib_worker_1.6.2`, can we combine them into one jar file: libWorker1.6.2.jar : https://stackoverflow.com/questions/5080220/how-to-combine-two-jar-files

amamounelsayed:
> This PR includes individual jar files in folder `lib_worker_1.6.2`, can we combine them into one jar file: libWorker1.6.2.jar : https://stackoverflow.com/questions/5080220/how-to-combine-two-jar-files

Got it, this may add more complication to the jar loading. I can have a look but this approach is just temporary solution.

pragnagopa:
> Got it, this may add more complication to the jar loading. I can have a look but this approach is just temporary solution.

I missed mentioning this before, can you please update PR description with size of the java worker folder with this change? If it is significantly higher, it is worth combining jars and checking the size. Due to size restrictions on windows VMs, we need to ensure worker size does not change much with this change.

amamounelsayed:
> I missed mentioning this before, can you please update PR description with size of the java worker folder with this change? If it is significantly higher, it is worth combining jars and checking the size. Due to size restrictions on windows VMs, we need to ensure worker size does not change much with this change.

The nuget will be increased by around 12MB.

pragnagopa:
> The nuget will be increased by around 12MB.

Thanks. Not a blocker then!

pragnagopa:
Does this PR address: #https://github.com/Azure/azure-functions-java-worker/issues/340 and #183 ?

amamounelsayed:
Thank you so much @pragnagopa for your time.

> Does this PR address: ##340 and #183 ?

Yeap this is correct also 
https://github.com/Azure/azure-functions-java-worker/issues/373
https://github.com/Azure/azure-functions-java-worker/issues/368
https://github.com/Azure/azure-functions-java-worker/issues/365
https://github.com/Azure/azure-functions-java-worker/issues/341

pragnagopa:
Updated the description to include references to issues to ensure merging PR closed all of them.

endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/HttpTriggerTests.java
TsuyoshiUshio: Add comment is fine.

package.ps1
TsuyoshiUshio: Sure!

PR #391, opened 2020-07-23
Bump the version to 1.7.2-SNAPSHOT

TsuyoshiUshio:


appveyor.yml
```diff
@@ -30,7 +30,7 @@ build_script:
        Write-Host “Found git tag."
       } 
       else {
-       $buildNumber = "1.7.1-$env:APPVEYOR_BUILD_NUMBER"
+       $buildNumber = "1.7.2-$env:APPVEYOR_BUILD_NUMBER"
```
amamounelsayed, line 33: We missed the SNAPSHOT

TsuyoshiUshio: Good catch

PR #402, opened 2020-09-17
CI Improvement

TsuyoshiUshio:
This PR is testing the Azure DevOps CI testing experiment.  It will require several round trip testing with Azure DevOps. 

The purpose is 

1. Simplify the CI pipeline
2. Easy to debug and see the test results easily

I start with CI testing part.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,6 +4,13 @@
 
 pool:
   vmImage: 'windows-2019'
+trigger:
+  branches:
+    include:
+      - dev
+      - master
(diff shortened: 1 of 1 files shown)


TsuyoshiUshio:
The initial investigation is done. As you can see, the test result is much easier to see it. 
I also solve the issue that sometimes the two CI runs at the same time. After coming back, I'll apply to the other tests, then this PR might be ready to merge. 

image

amamounelsayed (dismissed):

PR #411, opened 2020-11-23
Update for Distributed Tracing

TsuyoshiUshio:
This PR is for updating `worker.config` for enabling distributed tracing. 

I introduced `MESH_JAVA_OPTS` to the argument. 

At first, I used `JAVA_OPTS` however, some customers seem to use it. 
I introduce a new variables the is used only for mesh image configuration. 

Any naming suggesting is welcome.


worker.config.json
```diff
@@ -4,6 +4,6 @@
         "extensions": [".jar"],
         "defaultExecutablePath": "%JAVA_HOME%/bin/java",
         "defaultWorkerPath": "azure-functions-java-worker.jar",
-        "arguments": ["-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -noverify -Djava.net.preferIPv4Stack=true -jar", "%JAVA_OPTS%"]
+        "arguments": ["-XX:+TieredCompilation -XX:TieredStopAtLevel=1 -noverify -Djava.net.preferIPv4Stack=true -jar", "%JAVA_OPTS%", "%MESH_JAVA_OPTS%"]
```
amamounelsayed, line 7: Thank you sir, can we name I something related to AZURE_FUNCTIONS_MESH_JAVA_OPTS?
Also did we test this change with existing JAVA_OPTS and make sure that both will be picked with the host?

TsuyoshiUshio:
Hi @amamounelsayed 
I tested both on Windows and Mesh image.  I put a log (Linux) and Debug(Windows) 
on this part and find it works.  https://github.com/Azure/azure-functions-host/blob/0f18eade4b28b8cdde9e6a63dd0b7d4d0a051329/src/WebJobs.Script/Workers/ProcessManagement/DefaultWorkerProcessFactory.cs#L60

CI looks fails by CosmosDB that is not related this change. I'll re-run it.

TsuyoshiUshio:
It looks SNAT issue? 

```
System.Net.Http.HttpRequestException : No connection could be made because the target machine actively refused it
---- System.Net.Sockets.SocketException : No connection could be made because the target machine actively refused it
```

TsuyoshiUshio:
I run the pipeline one more time, If it won't fix, I'll make the client to static.

worker.config.json
TsuyoshiUshio: I've done that. I also finish testing both on windows and mesh images

TsuyoshiUshio:
I found the root cause of the CI fails. It looks reach the timeout when it reach the CosmosDB for reading a document. I can't see the CosmosDB resource from my subscription. Other socket exception is subsequent result of the timeout. So that I increase the timeout for solving it. 
https://azfunc.visualstudio.com/Azure%20Functions/_build/results?buildId=9802&view=ms.vss-test-web.build-test-results-tab&runId=1062300&resultId=100007&paneView=debug Could you have a look and approve it? I also change the http client to static. It is better solution, so that I keep it.

amamounelsayed (approved):

amamounelsayed:
@TsuyoshiUshio tested the Java_opts and new env AZURE_FUNCTIONS_MESH_JAVA_OPTS and both work together and do not override each others.

Issue #381, opened 2020-07-02
Azure Functions Java Worker 1.7.1 -- Dependencies change opt in


amamounelsayed:
This release is for shading all worker jars. This will give the option to clients to override the current behavior for how java worker jars are load it.

Background, before this release azure functions worker dependence takes precedence over the customer code dependencies in case they use the same ones. The current work is to give a way to customer dependencies to takes higher precedence by loading it first.

Only for JAVA 8
To be a backward compatible We freeze all the jars we currently have added to lib_worker_1.6.2.
We still load the worker jars first then the client jars.

If the client like to reverse this order and their jars takes presence first, they will need to add FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS True or 1. By this way it will be opt-in and load their jars first then worker jars.

JAVA 11 will use only the customer jars no need to add any application settings.

This will make java functions compatible with the latest spring and java azure sdks.

Fixes #340, Fixes #183, Fixes #373, Fixes #368, Fixes #365, Fixes #341

Releases contain the enhancement, 
v3 3.0.14251 or higher
v2 2.0.14248 or higher.

Core tools,
v3 3.0.2852 or higher
v2  2.7.2855 or higher

Windows deployment done.
Linux consumption deployment done.
Linux dedicated/premium, deployment done.
Core tools, released.

jongio:
@anirudhgarg @TsuyoshiUshio - Could you please provide an update on this?  Would be great to get it addressed as we have stale related issues that are dependent on this fix.  Thanks, Jon

amamounelsayed:
@jongio Thank you so much, We updated the status of the deployment. Please let us know if you have any questions.

amamounelsayed:
**Releases contain the enhancement,**
v3 3.0.14251 or higher
v2 2.0.14248 or higher.

**Windows** deployment done.
**Linux consumption** deployment done.
**Linux dedicated/premium**, still deploying.

jongio:
@g2vinay - Please have a look at previous comment and see if we can resolve ours with the current deployment.

amamounelsayed:
Releases contain the enhancement,
v3 3.0.14251 or higher
v2 2.0.14248 or higher.

Windows deployment done.
Linux consumption deployment done.
Linux dedicated/premium, deployment done.

amamounelsayed:
This fix is not released yet by core tools, it will be released in the next one.
[...]

PR #443, opened 2021-08-24
remove work lib logics

kaibocai:
Remove the logics for customer opt-in. Right now, customer is required to provide their own lib directory. 
Question to be clarified: one unit test failed due to the code change. Since this unit test didn't provide a it's lib directory. 
Proposed solution: copy the lib folder to target folder when build the package. This lib folder will work as a customer lib directory for unit tests.

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
@@ -89,11 +89,10 @@ public class JavaFunctionBroker {
 	private void addSearchPathsToClassLoader(FunctionMethodDescriptor function) throws IOException {
 		URL jarUrl = new File(function.getJarPath()).toURI().toURL();
 		classLoaderProvider.addUrl(jarUrl);
-		if(function.getLibDirectory().isPresent()) {
-			registerWithClassLoaderProvider(function.getLibDirectory().get());
-		} else {
-			registerWithClassLoaderProviderWorkerLibOnly();
+		if(!function.getLibDirectory().isPresent()) {
+			throw new IllegalArgumentException("Customer lib directory is not set");
 		}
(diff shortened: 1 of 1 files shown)


TsuyoshiUshio:
Hi @kaibocai
I did some research.  The spec will be like this. 

To give the context,  when we are Java8, we switched the order of the read 3rd party library and azure functions library by the environment variable ` FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS`. https://github.com/Azure/azure-functions-java-worker/issues/381

It is the only feature for Java8. So that we don't need it for V4. 
If the customer have a lib directory, load the customer libraries. If they don't have it, it's ok. 
https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-java?tabs=bash%2Cconsumption#third-party-libraries

So that you need to keep this logic. 
https://github.com/Azure/azure-functions-java-worker/pull/443/files#diff-8ba6922619e77013eed80700fc98dcbe05895acc55b1d705533074889f5528a0L92-L95

kaibocai:
> Hi @kaibocai
> I did some research. The spec will be like this.
> 
> To give the context, when we are Java8, we switched the order of the read 3rd party library and azure functions library by the environment variable ` FUNCTIONS_WORKER_JAVA_LOAD_APP_LIBS`. #381
> 
> It is the only feature for Java8. So that we don't need it for V4.
> If the customer have a lib directory, load the customer libraries. If they don't have it, it's ok.
> https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-java?tabs=bash%2Cconsumption#third-party-libraries
> 
> So that you need to keep this logic.
> https://github.com/Azure/azure-functions-java-worker/pull/443/files#diff-8ba6922619e77013eed80700fc98dcbe05895acc55b1d705533074889f5528a0L92-L95

Make sense. Thank you!

PR #444, opened 2021-08-25
work libs removed

kaibocai:
Remove the folder lib_worker_1.6.2 in java language worker. 
Modified the code so for java 8 it will use only the customer jars just like java 11.


TsuyoshiUshio:
Hi @kaibocai 
This change is fine, however, the target branch looks `dev`.  can we re-target to `v4` branch?

TsuyoshiUshio (approved):
It looks good to me. See the Unit/E2E testing result. 
This is ok for not including this PR, however, we'd better to remove the following part. https://github.com/Azure/azure-functions-java-worker/blob/v4/azure-pipelines.yml#L87-L105

TsuyoshiUshio:
Probably, we need to add `v4` branch in here for make v4 branch PR validation avaiable. 
https://github.com/Azure/azure-functions-java-worker/blob/v4/azure-pipelines.yml#L15

Francisco-Gamino:
@TsuyoshiUshio / @kaibocai -- Since the `dev` branch is always pointing to the latest working branch, should `dev` now point to `v4`? If this is the case, then the change should target the `dev` branch.

TsuyoshiUshio:
Hi @Francisco-Gamino 
Let me share the context. This PR is for the java worker for v4.  However, we might do some change against dev (that means v3) So that keep it as is for a while, at least until Ahmed is back.

endtoendtests/Azure.Functions.Java.Tests.E2E/Azure.Functions.Java.Tests.E2E/HttpEndToEndTests.cs
```diff
@@ -42,22 +42,9 @@ public async Task HttpTrigger_ReturnsCustomCode()
...
-            if (java_home.Contains("zulu11") || (value != null && value.ToLower().Equals("true")))
-            {
-                Assert.True(await Utilities.InvokeHttpTrigger("HttpTriggerJavaClassLoader", "?&name=Test", HttpStatusCode.OK, "Test"));
-            }
-            else
-            {
-                Assert.True(await Utilities.InvokeHttpTrigger("HttpTriggerJavaClassLoader", "?&name=Test", HttpStatusCode.InternalServerError, ""));
-            }
+
+            Assert.True(await Utilities.InvokeHttpTrigger("HttpTriggerJavaClassLoader", "?&name=Test", HttpStatusCode.OK, "Test"));
```
kaibocai, line 46: @TsuyoshiUshio I change the test cases here, before for java 8 if work lib is not found it will throw exception, but after the change this case should also works fine.

TsuyoshiUshio: Hi @kaibocai 
Looks good and thank you for noticing a lot of things and remove some constants. 
Ready for merge for v4 branch. After Ahmed is coming back, Let's test with integration testing.

PR #447, opened 2021-08-27
fix function name issue

kaibocai:
Fix the issue that function cannot have same method name as any other method in it's parent class.


TsuyoshiUshio (commented):
Thank you for finding out this issue. 
For the change, Looks good to me. However, The CI fails by the other reason. Could you find the reason why it fails? 
Once CI has passed, I'd happy to approve this change.

kaibocai:
> Thank you for finding out this issue.
> For the change, Looks good to me. However, The CI fails by the other reason. Could you find the reason why it fails?
> Once CI has passed, I'd happy to approve this change.

Hi @TsuyoshiUshio, yes currently trying to identify the reason and will get it resolved. Thanks.

PR #460, opened 2021-09-22
Use a single classloader for java 11

### LINKED ISSUE #412: Java11: Static variable appears to be difference instance even though it's in the same class.
#### Repro steps

Provide the steps required to reproduce the problem:

1. Create Function.java below.
```
package com.azure.yutobo.func;

import com.microsoft.azure.functions.ExecutionContext;
import com.microsoft.azure.functions.HttpMethod;
import com.microsoft.azure.functions.HttpRequestMessage;
import com.microsoft.azure.functions.HttpResponseMessage;
import com.microsoft.azure.functions.HttpStatus;
import com.microsoft.azure.functions.annotation.AuthorizationLevel;
import com.microsoft.azure.functions.annotation.FunctionName;
import com.microsoft.azure.functions.annotation.HttpTrigger;

import java.util.Optional;

public class Function {
    static final Person person  = new Person();

    @FunctionName("foo")
    public HttpResponseMessage foo(
            @HttpTrigger(
                name = "req",
                methods = {HttpMethod.GET, HttpMethod.POST},
                authLevel = AuthorizationLevel.ANONYMOUS)
                HttpRequestMessage<Optional<String>> request,
        context.getLogger().info("foo processed a request.");
        return request.createResponseBuilder(HttpStatus.OK).body("Foo: " + person.).build();
    }

    @FunctionName("bar")
    public HttpResponseMessage bar(
            @HttpTrigger(
                name = "req",
                methods = {HttpMethod.GET, HttpMethod.POST},
                authLevel = AuthorizationLevel.ANONYMOUS)
                HttpRequestMessage<Optional<String>> request,
        context.getLogger().info("bar processed a request.");
        return request.createResponseBuilder(HttpStatus.OK).body("Bar: " + person).build();
    }
}
```
```
package com.azure.yutobo.func;

class Person {
    String name = "";
    int age = 18;

    Person() {}

    @Override
    public int hashCode() {
        final int prime = 31;
        int result = 1;
        result = prime * result + age;
        result = prime * result + ((name == null) ? 0 : name.hashCode());
        return result;
    }

    @Override
    public boolean equals(Object obj) {
        if (this == obj)
            return true;
        if (obj == null)
            return false;
        if (getClass() != obj.getClass())
            return false;
        Person other = (Person) obj;
        if (age != other.age)
            return false;
        if (name == null) {
            if (other.name != null)
                return false;
        } else if (!name.equals(other.name))
            return false;
        return true;
    }

    @Override
    public String toString() {
        return "Person [age=" + age + ", name=" + name + "]";
    }
}

```

2. access Foo & Bar endpoints.

#### Expected behavior

Foo & Bar methods show the same output because they share the same class variable.

#### Actual behavior

Foo & Bar methods show the different output like below.
image

I found that there are separate class variables of Function.class in the heap.
image

And, Function.class loaded twince by the different URLClassLoader.
Apparently, you don't check whether Function.class is already loaded or not.

#### Known workarounds
I couldn't find workarounds.

#### Impact
- There appears to be no ways to share the class variable even methods are in the same class.
- I think we can't make Double-checked locked instance to get Singleton in multi-thread environment.

shreyas-gopalakrishna:
Fixes
- For java 11 a synchronized singleton classloader will be used
- This fixes the issue - multiple instances of the static variable even though it's in the same class

Code changes
- Creating a singleton of the URLClassLoader which will be reused as necessary. 
- In order to handle synchronization across threads, the double checked locking with singleton is done.
- Using the volatile keyword to avoid any cache incoherence issues
- Test cases are updated to check if static variables are updated under the same instance

#412

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
@@ -19,12 +19,18 @@ public class EnhancedClassLoaderProvider implements ClassLoaderProvider {
      */
     @Override
     public ClassLoader createClassLoader() {
-        URL[] urlsForClassLoader = new URL[urls.size()];
-        urls.toArray(urlsForClassLoader);
-
-        URLClassLoader classLoader = new URLClassLoader(urlsForClassLoader);
-        loadDrivers(classLoader);
-        return classLoader;
+        if (classLoaderInstance == null) {
+            synchronized (lock) {
+                if (classLoaderInstance == null) {
+                    URL[] urlsForClassLoader = new URL[urls.size()];
+                    urls.toArray(urlsForClassLoader);
+                    URLClassLoader loader = new URLClassLoader(urlsForClassLoader);
+                    loadDrivers(loader);
+                    classLoaderInstance = loader;
+                }
+            }
+        }
+        return classLoaderInstance;
     }
 
     private void loadDrivers(URLClassLoader classLoader) {
@@ -73,4 +79,6 @@ public class EnhancedClassLoaderProvider implements ClassLoaderProvider {
         urls.add(url);
     }
     private final Set<URL> urls;
+    private final Object lock = new Object();
+    private static volatile URLClassLoader classLoaderInstance;
 }
--- a/endtoendtests/Azure.Functions.Java.Tests.E2E/Azure.Functions.Java.Tests.E2E/HttpEndToEndTests.cs
+++ b/endtoendtests/Azure.Functions.Java.Tests.E2E/Azure.Functions.Java.Tests.E2E/HttpEndToEndTests.cs
@@ -46,5 +46,14 @@ namespace Azure.Functions.Java.Tests.E2E
             Assert.True(await Utilities.InvokeHttpTrigger("HttpTriggerJavaClassLoader", "?&name=Test", HttpStatusCode.OK, "Test"));
 
         }
+
+        [Fact]
+        public async void HttpTriggerJavaStatic()
+        {
+            await HttpTriggerTests("HttpTriggerJavaStatic1", "", HttpStatusCode.OK, "1");
+            await HttpTriggerTests("HttpTriggerJavaStatic2", "", HttpStatusCode.OK, "2");
+            await HttpTriggerTests("HttpTriggerJavaStatic1", "", HttpStatusCode.OK, "3");
+            await HttpTriggerTests("HttpTriggerJavaStatic2", "", HttpStatusCode.OK, "4");
+        }
     }
 }
--- a/endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/HttpTriggerTests.java
+++ b/endtoendtests/src/main/java/com/microsoft/azure/functions/endtoend/HttpTriggerTests.java
@@ -237,4 +237,32 @@ public class HttpTriggerTests {
             return request.createResponseBuilder(HttpStatus.OK).body("Hello, " + name).build();
         }
     }
+
+    private static int flag = 0;
+
+    @FunctionName("HttpTriggerJavaStatic1")
+    public HttpResponseMessage HttpTriggerJavaStatic1(
+            @HttpTrigger(
+                    name = "req",
+                    methods = {HttpMethod.GET, HttpMethod.POST},
+                    authLevel = AuthorizationLevel.ANONYMOUS)
+                    HttpRequestMessage<Optional<String>> request,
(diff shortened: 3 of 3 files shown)


src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -73,4 +88,5 @@ public void addUrl(URL url) throws IOException {
         urls.add(url);
     }
     private final Set<URL> urls;
+    private static volatile URLClassLoader classLoaderInstance;
```
shreyas-gopalakrishna, line 83: Using the volatile keyword to avoid any cache incoherence issues. Review if this is necessary for our use case

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -73,4 +88,5 @@ public void addUrl(URL url) throws IOException {
         urls.add(url);
     }
     private final Set<URL> urls;
```
trask, line 81:
```suggestion
    private final Set<URL> urls;
    private final Object lock = new Object();
```

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -22,11 +22,26 @@ public ClassLoader createClassLoader() {
...
         return classLoader;
     }
 
+    /**
+     * @param urlsForClassLoader Array of URLs to construct a new URLClassLoader
+     * @return single instance of URLClassLoader
+     */
+    private URLClassLoader getURLClassLoaderInstance(URL[] urlsForClassLoader) {
+        if (classLoaderInstance == null) {
+            synchronized (URLClassLoader.class) {
```
trask:
```suggestion
            synchronized (lock) {
```

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -22,11 +22,26 @@ public ClassLoader createClassLoader() {
...
-        loadDrivers(classLoader);
+        URLClassLoader classLoader = getURLClassLoaderInstance(urlsForClassLoader);
         return classLoader;
     }
 
+    /**
+     * @param urlsForClassLoader Array of URLs to construct a new URLClassLoader
+     * @return single instance of URLClassLoader
+     */
+    private URLClassLoader getURLClassLoaderInstance(URL[] urlsForClassLoader) {
```
trask: no need to pass args here, calculate below instead

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -19,12 +19,25 @@ public EnhancedClassLoaderProvider() {
      */
     @Override
     public ClassLoader createClassLoader() {
-        URL[] urlsForClassLoader = new URL[urls.size()];
-        urls.toArray(urlsForClassLoader);
+        return createURLClassLoaderInstance();
```
trask: just a style thought, can inline the `createURLClassLoaderInstance` method here

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -19,12 +19,25 @@ public EnhancedClassLoaderProvider() {
...
+     * @return instance of URLClassLoader
+     */
+    private URLClassLoader createURLClassLoaderInstance() {
+        if (classLoaderInstance == null) {
+            synchronized (lock) {
+                if (classLoaderInstance == null) {
+                    URL[] urlsForClassLoader = new URL[urls.size()];
+                    urls.toArray(urlsForClassLoader);
+                    classLoaderInstance = new URLClassLoader(urlsForClassLoader);
+                    loadDrivers(classLoaderInstance);
```
trask: good to assign the field at the very end, so that another thread can't come in and return `classLoaderInstance` prior to it being fully initialized
```suggestion
                    URLClassLoader loader = new URLClassLoader(urlsForClassLoader);
                    loadDrivers(loader);
                    classLoaderInstance = loader;
```

trask (dismissed):

amamounelsayed (dismissed):

trask (approved):

amamounelsayed (approved):

PR #461, opened 2021-09-22
Use a single classloader for java 11 if environment variable is set

### LINKED ISSUE #412: Java11: Static variable appears to be difference instance even though it's in the same class.

### LINKED ISSUE #465: Java11: v3 - Static variable appears to be in different instance even though it's in the same class.
#### Investigative information

Using a static variable in java 11 appears to be treated as different instance even though it exists in the same class. 
The issue is similar to #412 which will be fixed in v3 with an application setting. 

#### Repro steps

Use a static variable in a Java function and update it from multiple objects.

#### Expected behavior

The variable must be static across all objects. 

#### Actual behavior

The variable appears to be in different instance and updated individually in every instance.

#### Related information 
This issue will be fixed in v3 provided `FUNCTIONS_WORKER_JAVA_V3_SINGLE_CLASSLOADER` application setting/environment variable is set to true. If the variable is set to false or if left unset, the original behavior is retained.

This issue will be a break change in v4 with no application setting/environment variable need and updates on it can be found at https://github.com/Azure/Azure-Functions/issues/1997

shreyas-gopalakrishna:
If the `FUNCTIONS_WORKER_JAVA_SINGLE_CLASSLOADER` environment variable is set to true then the following changes are made
- For java 11 a synchronized singleton classloader is created for usage 
- This fixes the issue - multiple instances of the static variable even though it's in the same class

#412

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
@@ -19,9 +19,31 @@ public class EnhancedClassLoaderProvider implements ClassLoaderProvider {
      */
     @Override
     public ClassLoader createClassLoader() {
+        if(Helper.isCustomURLClassLoader()) {
+            return getURLClassLoaderInstance();
+        }else {
+            return createURLClassLoaderInstance();
+        }
+    }
+
+    /**
+     * Create and return a singleton URL classloader
+     * @return instance of URLClassLoader
+     */
+    private URLClassLoader getURLClassLoaderInstance() {
+        if (classLoaderInstance == null) {
+            synchronized (lock) {
+                if (classLoaderInstance == null) {
+                    classLoaderInstance = createURLClassLoaderInstance();
+                }
+            }
+        }
+        return classLoaderInstance;
+    }
+
+    private URLClassLoader createURLClassLoaderInstance(){
         URL[] urlsForClassLoader = new URL[urls.size()];
         urls.toArray(urlsForClassLoader);
-
         URLClassLoader classLoader = new URLClassLoader(urlsForClassLoader);
         loadDrivers(classLoader);
         return classLoader;
@@ -73,4 +95,6 @@ public class EnhancedClassLoaderProvider implements ClassLoaderProvider {
         urls.add(url);
     }
     private final Set<URL> urls;
+    private final Object lock = new Object();
+    private static volatile URLClassLoader classLoaderInstance;
 }
--- a/endtoendtests/Azure.Functions.Java.Tests.E2E/Azure.Functions.Java.Tests.E2E/HttpEndToEndTests.cs
+++ b/endtoendtests/Azure.Functions.Java.Tests.E2E/Azure.Functions.Java.Tests.E2E/HttpEndToEndTests.cs
@@ -59,5 +59,19 @@ namespace Azure.Functions.Java.Tests.E2E
                 Assert.True(await Utilities.InvokeHttpTrigger("HttpTriggerJavaClassLoader", "?&name=Test", HttpStatusCode.InternalServerError, ""));
             }
         }
+
+        [Fact]
+        public async void HttpTriggerJavaStatic()
+        {
+            String value = Environment.GetEnvironmentVariable("FUNCTIONS_WORKER_JAVA_V3_SINGLE_CLASSLOADER");
+            String java_home = Environment.GetEnvironmentVariable("JAVA_HOME");
+            if (java_home.Contains("1.8") || (value != null && (value.ToLower().Equals("true") || value.ToLower().Equals("1"))))
+            {
+                await HttpTriggerTests("HttpTriggerJavaStatic1", "", HttpStatusCode.OK, "1");
+                await HttpTriggerTests("HttpTriggerJavaStatic2", "", HttpStatusCode.OK, "2");
+                await HttpTriggerTests("HttpTriggerJavaStatic1", "", HttpStatusCode.OK, "3");
+                await HttpTriggerTests("HttpTriggerJavaStatic2", "", HttpStatusCode.OK, "4");
+            }
+        }
     }
 }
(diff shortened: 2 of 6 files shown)


src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -22,11 +22,32 @@ public ClassLoader createClassLoader() {
...
+    /**
+     * @param urlsForClassLoader Array of URLs to construct a new URLClassLoader
+     * @return single instance of URLClassLoader
+     */
+    private URLClassLoader getURLClassLoaderInstance(URL[] urlsForClassLoader) {
+        if (classLoaderInstance == null) {
+            synchronized (URLClassLoader.class) {
+                if (classLoaderInstance == null) {
+                    classLoaderInstance = new URLClassLoader(urlsForClassLoader);
+                    loadDrivers(classLoaderInstance);
```
trask: extract to function to reuse from above also

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -19,12 +19,36 @@ public EnhancedClassLoaderProvider() {
...
+    /**
+     * Create and return a singleton URL classloader
+     * @return instance of URLClassLoader
+     */
+    private URLClassLoader getURLClassLoaderInstance() {
+        if (classLoaderInstance == null) {
+            synchronized (lock) {
+                if (classLoaderInstance == null) {
+                    URLClassLoader loader = createURLClassLoaderInstance();
+                    loadDrivers(loader);
```
trask: looks like `loadDrivers` can be moved to `createURLClassLoaderInstance`

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -19,9 +19,33 @@ public EnhancedClassLoaderProvider() {
      */
     @Override
     public ClassLoader createClassLoader() {
+        if(Helper.isCustomURLClassLoader()) {
+            return getURLClassLoaderInstance();
+        }else{
+            URLClassLoader classLoader = createURLClassLoaderInstance();
+            return classLoader;
```
trask:
```suggestion
            return createURLClassLoaderInstance();
```

src/main/java/com/microsoft/azure/functions/worker/reflect/EnhancedClassLoaderProvider.java
```diff
@@ -19,9 +19,33 @@ public EnhancedClassLoaderProvider() {
...
+    /**
+     * Create and return a singleton URL classloader
+     * @return instance of URLClassLoader
+     */
+    private URLClassLoader getURLClassLoaderInstance() {
+        if (classLoaderInstance == null) {
+            synchronized (lock) {
+                if (classLoaderInstance == null) {
+                    URLClassLoader loader = createURLClassLoaderInstance();
+                    classLoaderInstance = loader;
```
trask:
```suggestion
                    classLoaderInstance = createURLClassLoaderInstance();
```

trask (dismissed):
just two minor simplifications

trask (dismissed):

amamounelsayed (approved):

PR #466, opened 2021-09-29
Release 4.0.0

amamounelsayed:
Release 4.0.0 has the following changes:
- For Java 11 - usage of a classloader singleton which is shared across the java worker 
- Remove the folder lib_worker_1.6.2 in java language worker. Java 8 it will not have default worker libraries

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,7 +45,7 @@ jobs:
             Write-Host "Found git tag."
           }
           else {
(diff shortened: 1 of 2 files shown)


kaibocai:
This release includes two break changes:

For java 11 a synchronized singleton classloader will be used. This fixes the issue Azure/Azure-Functions#1997
Java 8 stop supporting loading the worker jars, working jars are removed from java worker. Azure/Azure-Functions#1991

PR #469, opened 2021-09-30
Updated pom.xml with grpc netty shaded

shreyas-gopalakrishna:
- Updated dependency from grpc-netty to use grpc-netty-shaded
- Adjusted the shading logic for grpc-netty-shaded
- This fixes Component Governance from picking up alerts from grpc-netty modules we don't directly use

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
 		</dependency>
 		<dependency>
 			<groupId>io.grpc</groupId>
-			<artifactId>grpc-netty</artifactId>
+			<artifactId>grpc-netty-shaded</artifactId>
 			<version>1.20.0</version>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -247,6 +247,16 @@
 								<relocation combine.children="append">
 									<pattern>io.grpc</pattern>
 									<shadedPattern>com.microsoft.azure.functions.shaded.io.grpc</shadedPattern>
+									<excludes>
+										<exclude>io.grpc.netty.shaded.io.grpc.netty.*</exclude>
+									</excludes>
+								</relocation>
+								<relocation>
+									<pattern>io.grpc.netty.shaded.io.grpc.netty</pattern>
+									<shadedPattern>com.microsoft.azure.functions.shaded.io.grpc.netty.shaded.io.grpc.netty</shadedPattern>
```
amamounelsayed, line 256: Do we see any issue with length of the package with different OS?

trask: no this should be fine 👍

trask (approved):

amamounelsayed (approved):

PR #483, opened 2021-10-27
update core tools and .net version for v4 end2end testing

kaibocai:
update core tools to version 4 for end2end testing.


setup-tests.ps1
```diff
@@ -1,57 +0,0 @@
-
```
Francisco-Gamino, line 1: I was not aware this code is a dup of `setup-tests-pipeline.ps1`. Are we referencing `setup-tests.ps1` anywhere?

Francisco-Gamino (dismissed):
One minor comment. Otherwise, LGTM.

setup-tests.ps1
kaibocai: It was referenced in build.ps1 which is not used anywhere in the project. I believe they are unused files (build.ps1 and setup-tests.ps1). @amamounelsayed please confirm it. Thank you.

azure-pipelines.yml
```diff
@@ -57,6 +57,17 @@ jobs:
           TargetFolder: '$(Build.ArtifactStagingDirectory)'
           CleanTargetFolder: true
         displayName: 'Copying files for artifacts'
+
+      - pwsh: |
+          Invoke-WebRequest 'https://raw.githubusercontent.com/dotnet/cli/master/scripts/obtain/dotnet-install.ps1' -OutFile 'dotnet-install.ps1'
+          ./dotnet-install.ps1 -InstallDir "$env:ProgramFiles/dotnet" -Version "6.0.100-preview.7.21379.14" -Channel 'release'
```
Francisco-Gamino: Please update this version to be `6.0.100-rc.2.21505.57`. This is what the Core Tools is currently using https://github.com/Azure/azure-functions-core-tools/blob/v4.x/azure-pipelines.yml#L49.

kaibocai: updated it.

Francisco-Gamino: Great, thanks.

PR #537, opened 2022-03-11
add logics support return value for tirggers like durable functions

kaibocai:

add logics support return value for tirggers like durable functions, match with https://github.com/Azure/azure-functions-java-library/pull/171
### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/broker/CoreTypeResolver.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/broker/CoreTypeResolver.java
@@ -4,6 +4,7 @@ import java.lang.annotation.*;
 import java.lang.reflect.*;
 
 import com.microsoft.azure.functions.*;
+import com.microsoft.azure.functions.worker.Constants;
 
 public class CoreTypeResolver {
 	private static boolean isOutputParameter(Type target) {
@@ -119,4 +120,22 @@ public class CoreTypeResolver {
 		}
 		return new String("");
 	}
+
+    static boolean checkHasImplicitOutput(Parameter parameter){
+        try {
+            Annotation[] annotations = parameter.getAnnotations();
+            for (Annotation annotation : annotations) {
+				String packageName = annotation.annotationType().getPackage().getName();
+				if (!packageName.contains(Constants.JAVA_LIBRARY_PACKAGE_PREFIX) || !packageName.endsWith(Constants.JAVA_LIBRARY_PACKAGE_SUFFIX)) continue;
+                Method hasImplicitOutput = annotation.annotationType().getMethod("hasImplicitOutput");
+                if (hasImplicitOutput == null) continue;
+                boolean isImplicitOutput = (boolean) hasImplicitOutput.invoke(annotation);
+                if (isImplicitOutput) return true;
+            }
+            return false;
+        } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+            //do nothing
+        }
+        return false;
+    }
 }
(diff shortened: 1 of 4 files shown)


cgillum (dismissed):

src/main/java/com/microsoft/azure/functions/worker/broker/CoreTypeResolver.java
```diff
@@ -119,4 +120,21 @@ static String getBindingNameAnnotation(Parameter param) {
 		}
 		return new String("");
 	}
+
+    static boolean checkHasImplicitOutput(Parameter parameter){
+        try {
+            Annotation[] annotations = parameter.getAnnotations();
+            for (Annotation annotation : annotations) {
+				if (!annotation.annotationType().getName().contains(Constants.JAVA_LIBRARY_PACKAGE_PREFIX)) continue;
```
kaibocai, line 128: this will return class qualified name which ends with Class name, I think only check Prefix that match with azure functions is enough. If we insist to check it package name end with annotation it requires string processing, it may affect the performance as well and involving potential bugs.

src/main/java/com/microsoft/azure/functions/worker/broker/CoreTypeResolver.java
```diff
@@ -119,4 +120,22 @@ static String getBindingNameAnnotation(Parameter param) {
 		}
 		return new String("");
 	}
+
+    static boolean checkHasImplicitOutput(Parameter parameter){
+        try {
+            Annotation[] annotations = parameter.getAnnotations();
+            for (Annotation annotation : annotations) {
+				String packageName = annotation.annotationType().getPackage().getName();
+				if (!packageName.contains(Constants.JAVA_LIBRARY_PACKAGE_PREFIX) || !packageName.endsWith(Constants.JAVA_LIBRARY_PACKAGE_SUFFIX)) continue;
```
cgillum: Why do we restrict this feature to only annotations in the Java library?

amamounelsayed: Good question, for two reasons, 
- For performance, we do not like to iterate on all annotations the customers may add.
- We were worry that any of the customer annotations may have hasImplicitOutput annotation attribute for any other reason.

Thank you so much

cgillum: It looks like we're already iterating on all the annotations that the customer has, so I assume the performance benefit of this is pretty minimal.

As for not interfering with existing customer annotations, would it make sense then to instead have an annotation for trigger annotations rather than having a property/method and using reflection to read it?  For example, suppose we define a `@HasImplicitOutput` annotation that developers put on their trigger annotations. Here's what my orchestration trigger definition might look like in this proposal:

```java
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.PARAMETER)
@CustomBinding(direction = "in", name = "", type = "orchestrationTrigger")
@HasImplicitOutput(true)
public @interface DurableOrchestrationTrigger {
    // ...
}
```

By defining our own annotation type, we solve the problem of overlapping with something the customer has already defined.  It also removes the need for us to limit what trigger bindings can use implicit outputs. And as I mentioned already, we can remove reflection methods like `getMethod("hasImplicitOutput")`.

My fear is that we're unnecessarily limiting who can use the implicit output feature with the current design. Ideally it would work for both 1st party and 3rd party annotations. I'm not a Java expert but I think my proposed design enables this.

cgillum (approved):
This design looks good to me!

amamounelsayed (approved):

shreyas-gopalakrishna (approved):

PR #544, opened 2022-03-25
supporting java library restructure V3

kaibocai:

supporting java library restructure V3

### Pull request checklist

### Additional information

Additional PR information


src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -132,21 +135,53 @@ void registerWithClassLoaderProvider(File libDirectory) {
 
 				if(Helper.isLoadAppLibsFirst()) {
 					// load client app jars first.
-					classLoaderProvider.addDirectory(libDirectory);
-					classLoaderProvider.addDirectory(workerLib);
+					addDirectory(libDirectory, false);
```
kaibocai, line 138: Java8:
1. load cx first with annotation lib  ---> cx annotation lib will be put in cxset, worker annotation lib will be put in workerset, order is expected
2. load cx first without annotation lib --> no cx annotation lib in cxset, worker annotation lib will be put in workerset, as expected
3. load worker first with annotation lib --> cx annotation lib in cxset, worker annotation lib in workerset, order as expected
4. load worker first without annotation lib --> no cx annotation lib in cxset, worker annotation lib will be put in workerset, as expected

Java11: should be same as v4

amamounelsayed (approved):

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -132,21 +135,53 @@ void registerWithClassLoaderProvider(File libDirectory) {
...
+					addDirectory(workerLib, true);
 				} else {
 					// Default load worker jars first.
-					classLoaderProvider.addDirectory(workerLib);
-					classLoaderProvider.addDirectory(libDirectory);
+					addDirectory(workerLib, false);
+					addDirectory(libDirectory, true);
```
amamounelsayed, line 143: I believe this should be false, and line 142 is true, and we can do the same with 138 and 139.
The idea should be catch all in case the code does not contain the lib jar.

Please correct me if I am wrong

kaibocai: the flag here is used to avoid loading the lib jar in annotationlib folder inside worker twice. Also the name of the flag is ```skipJavaAnnotation```, if it's true it will skip the method load worker annotation jar, if it's false it will load. So we load it at the first time by setting the flag to false, and avoid load it again by setting the flag to true.

shreyas-gopalakrishna (approved):
LGTM!

PR #590, opened 2022-06-21
Release Java worker 2.3.0

shreyas-gopalakrishna:


pom.xml
```diff
@@ -25,7 +25,7 @@
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<azure.functions.java.core.library.version>1.0.0</azure.functions.java.core.library.version>
-		<azure.functions.java.library.version>2.0.1-SNAPSHOT</azure.functions.java.library.version>
+		<azure.functions.java.library.version>2.0.1</azure.functions.java.library.version>
```
kaibocai, line 28: should we also update the library jar file in annotationlib?

kaibocai (approved):

pom.xml
shreyas-gopalakrishna: Keeping the same jar for now. We will be updating it during next break change on library side

PR #631, opened 2022-08-18
Gracefully shutdown worker on WorkerTerminate request

### LINKED ISSUE #630: Allow java workers to terminate gracefully when WorkerTerminate message received
Host will send a [WorkerTerminate](https://github.com/Azure/azure-functions-language-worker-protobuf/blob/58b3dc0dbde065b15bd979bd2e27f36b7ee6273e/src/proto/FunctionRpc.proto#L39) message to worker to allow the worker to terminate itself gracefully instead of immediately killing it. Worker to act on this signal and send Success response to Host.

Related issue on Host side - https://github.com/Azure/azure-functions-host/issues/2308

shreyas-gopalakrishna:

resolves #630  https://github.com/Azure/azure-functions-host/issues/2308 https://github.com/Azure/azure-functions-host/issues/3296

attached design doc: https://dev.azure.com/msazure/One/_git/AAPT-Antares-Docs?path=/TeamDocs/FunctionTeamDocs/Design/OOPWorkers/GracefulShutdownOfWorker.md&_a=preview&anchor=out-of-proc-support-for-cancellation-tokens

The flow would be the following:

- If the worker does not support termination notification (based on capabilities RpcWorkerConstants.HandlesWorkerTerminateMessage), no change in the behavior.
- Otherwise; the host will send a termination message and wait for the process to exit up to the timeout for process termination.
- The worker's responsibility would be to handle that message and exit the process within the allotted time. In the case of .NET, application lifetime services will be notified and if customers are subscribing to shutdown events and tokens, they would also receive the notification. This would also terminate hosted services.
- If the timeout is hit, the process would be killed.

### Pull request checklist

  * [x] Otherwise: I've added my notes to `release_notes.md`


kaibocai:
Hi @liliankasem , just to confirm that host will restart after it send out WorkerTerminate request to language worker right. Do we long how many time we got between WorkerTerminate request send out and host restart? Thanks.

liliankasem:
> Hi @liliankasem , just to confirm that host will restart after it send out WorkerTerminate request to language worker right. Do we long how many time we got between WorkerTerminate request send out and host restart? Thanks.

Grace period is set [here](https://github.com/Azure/azure-functions-host/blob/26e4ad464062f87bc4864b0f49bf4739c00e1ffe/src/WebJobs.Script.Grpc/Channel/GrpcWorkerChannel.cs#L942), which I believe is currently 5 seconds @surgupta-msft can confirm

src/main/java/com/microsoft/azure/functions/worker/handler/WorkerInitRequestHandler.java
```diff
@@ -24,6 +24,7 @@ String execute(WorkerInitRequest request, WorkerInitResponse.Builder response) {
         response.putCapabilities("WorkerStatus", "WorkerStatus");
         response.putCapabilities("RpcHttpBodyOnly", "RpcHttpBodyOnly");
         response.putCapabilities("RpcHttpTriggerMetadataRemoved", "RpcHttpTriggerMetadataRemoved");
+        response.putCapabilities("HandlesWorkerTerminateMessage", "HandlesWorkerTerminateMessage");
```
kaibocai, line 27: Should the value be "true", can we confirm with host team

src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -122,6 +123,9 @@ public static void main(String[] args) {
         } else {
             try (JavaWorkerClient client = new JavaWorkerClient(app)) {
                 client.listen(app.getWorkerId(), app.getRequestId()).get();
+            } catch (GracefulTerminationException ex) {
```
kaibocai: Hi Shreyas, I don't think the exception will be catch as it will already be caught here https://github.com/Azure/azure-functions-java-worker/blob/ac6cb815d9493a0fcef7dd87a3006da3623b8b18/src/main/java/com/microsoft/azure/functions/worker/handler/MessageHandler.java#L49 right.

src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -122,6 +123,9 @@ public static void main(String[] args) {
...
+                WorkerLogManager.getSystemLogger().log(Level.INFO, "Gracefully shutting down the worker.");
+                System.exit(0);
```
kaibocai: before the exit, we may need to close the rpc channel and also maybe flush the log buffer, etc.

kaibocai (commented):
add some minor comments.

src/main/java/com/microsoft/azure/functions/worker/handler/WorkerInitRequestHandler.java
liliankasem: We only check if it's not null or empty, so value (as long as not empty) doesn't matter

src/main/java/com/microsoft/azure/functions/worker/WorkerLogManager.java
```diff
@@ -28,20 +29,33 @@ private void initializeImpl(JavaWorkerClient client, boolean logToConsole) {
...
-    private void deinitializempl() {
+    private void deinitializImpl() {
         assert this.client != null;
         clearHandlers(this.hostLogger);
         this.logToConsole = false;
         this.client = null;
     }
 
+    private void flushLogsImpl() {
+        flush(emptyLogger);
```
kaibocai: we may can remove flush this emptyLogger as it was not used anywhere

src/main/java/com/microsoft/azure/functions/worker/WorkerLogManager.java
```diff
@@ -28,20 +29,33 @@ private void initializeImpl(JavaWorkerClient client, boolean logToConsole) {
...
+        flush(systemLogger);
+        flush(hostLogger);
+    }
+
+    private static void flush(Logger logger) {
+        Handler[] handlers = logger.getHandlers();
```
kaibocai, line 48: it shouldn't have any issue, just want to double check for emptyLogger it won't have issue like NPE.

kaibocai (dismissed):
LGTM!

src/main/java/com/microsoft/azure/functions/worker/WorkerLogManager.java
shreyas-gopalakrishna: Yes, this can be removed

src/main/java/com/microsoft/azure/functions/worker/WorkerLogManager.java
shreyas-gopalakrishna: No, verified this will not throw a null pointer

kaibocai (approved):
👍

PR #638, opened 2022-08-30
Read files to avoid disk read during specialization

### LINKED ISSUE #637: Avoid reading from DISK during cold start specialization
When running a profiler for the placeholder pools, it is observed that there are some files which is being read from the disk during specialization.
Reading those files before specialization will help in paging in bytes to reduce the disk reads.

The java worker reads files such as below during specialization which impacts cold start numbers.
- C:\\Program Files\\Java\\microsoft-jdk-11.0.14.1\\bin\\server\\jvm.dll
- C:\Program Files\Java\microsoft-jdk-11.0.14.1\lib\modules

Read the files during start of the java worker to page in bytes so we avoid reading it during specialization

shreyas-gopalakrishna:
resolves #637 

### Pull request checklist

* [x] My changes **should not** be added to the release notes for the next release

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/Application.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/Application.java
@@ -1,10 +1,17 @@
 package com.microsoft.azure.functions.worker;
 
+import java.io.FileInputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Random;
 import java.util.logging.*;
 import javax.annotation.*;
 
 import org.apache.commons.cli.*;
 import org.apache.commons.lang3.exception.ExceptionUtils;
+
+import static com.microsoft.azure.functions.worker.Util.getJavaHome;
 
 /**
  * The entry point of the Java Language Worker. Every component could get the command line options from this singleton
@@ -115,6 +122,7 @@ public final class Application implements IApplication {
 
     public static void main(String[] args) {
         WorkerLogManager.getSystemLogger().log(Level.INFO, "Azure Functions Java Worker  version [ " + version() + "]");
+        readFilesForPlaceholders();
         Application app = new Application(args);
         if (!app.isCommandlineValid()) {
             app.printUsage();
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -129,6 +136,26 @@ public static void main(String[] args) {
...
+    private static void readFilesForPlaceholders() {
+        readFile(Paths.get(getJavaHome(), "lib", "modules"));
+        readFile(Paths.get(getJavaHome(), "bin", "server", "jvm.dll"));
+    }
+
+    private static void readFile(Path path) {
+        WorkerLogManager.getSystemLogger().log(Level.INFO, "Reading file " + path.toString() + " to page-in bytes.");
+        try {
+            if(Files.exists(path)) {
+                Files.readAllBytes(path);
```
shreyas-gopalakrishna: Revisit to check if this is safe to do or if we need to read file in chunks

kaibocai (dismissed):
LGTM!

src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -129,6 +136,26 @@ public static void main(String[] args) {
...
+            }
+        } catch (Exception ex) {
+            WorkerLogManager.getSystemLogger().log(Level.INFO, ExceptionUtils.getRootCauseMessage(ex), ex);
```
kaibocai, line 168: 👍

src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -129,6 +136,27 @@ public static void main(String[] args) {
         }
     }
 
+    /**
+     *  Read files during initialization to avoid disk reads during specialization. This is only to page-in bytes.
+     *  Example: jdk/lib/modules and jdk/bin/server/jvm.dll
+     */
+    private static void readFilesForPlaceholders() {
+        readFile(Paths.get(getJavaHome(), "lib", "modules"));
+        readFile(Paths.get(getJavaHome(), "lib", "rt.jar"));
```
trask:
```suggestion
        readFile(Paths.get(getJavaHome(), "lib", "modules")); // Java 11+
        readFile(Paths.get(getJavaHome(), "lib", "rt.jar")); // Java 8
```

src/main/java/com/microsoft/azure/functions/worker/Application.java
trask: I think good to read in chunks, otherwise will require a large continuous byte array

src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -129,6 +137,38 @@ public static void main(String[] args) {
...
     }
 
+    /**
+     * Read files during initialization to avoid disk reads during specialization. This is only to page-in bytes.
+     * Example: jdk/lib/modules and jdk/bin/server/jvm.dll
+     */
+    private static void readFilesForPlaceholders() {
+        readFile(Paths.get(getJavaHome(), "lib", "modules")); // Java 11+
+        readFile(Paths.get(getJavaHome(), "lib", "rt.jar")); // Java 8
+        readFile(Paths.get(getJavaHome(), "bin", "server", "jvm.dll"));
```
kaibocai, line 147: I believe this file only exists in windows, do we need to cache similar file for linux?

shreyas-gopalakrishna: Reading these files is decided based on the data in FunctionsColdStartAnalyzer. Currently we do not have this available for Linux. If we all agree this is needed, we can bring this change for Linux as well.

shreyas-gopalakrishna:
Closing this PR as reading these files is indirectly triggered when using the worker warmup functionality which addresses both Linux and windows.

PR #645, opened 2022-09-12
decouple the function definition and function executor.

kaibocai:

This PR mainly decoupled function definition (build at function load request processing) and function executor (build at invocation request process). It includes updates as below

- The worker used to build function executor at `FunctionLoadRequest`, now updated the logic to build the executor at `InvocationRequest`. 
- Update the executor only depends on the execution context passed to it at every invocation request, make the execution logic more clearly. 
- Making `ExecutionContextDataSource` class to `public`, as this class is only used internal in java worker, it should cause any issue on changing it's scope to `public`. 
- Update `ExecutionContextDataSource` to use builder pattern as it has to many arguments in old constructor. 
- Adding three fields `dataStore`, `methodBindInfo`, `containingClass` to `ExecutionContextDataSource`, since decoupled the function definition and function executor, now function executor relies on  `ExecutionContextDataSource` to get all required objects. 
- Update `MethodBindInfo`, `ParamBindInfo` from nested class to top class as they are used by other class besides their parent class.

Tried best to minimize the changes in this PR, all changes are necessary for the feature completeness. 

### Pull request checklist

* [ ] I have added all required tests (Unit tests, E2E tests) -- to be done. 

### Additional information

Additional PR information


kaibocai:
Open another similar pr without break change at https://github.com/Azure/azure-functions-java-worker/pull/652

PR #684, opened 2022-12-06
fix NPE issue

kaibocai:

resolves [#issue_for_this_pr](https://portal.microsofticm.com/imp/v3/incidents/details/353531000/home)

### Pull request checklist

* [X] I have added all required tests (Unit tests, E2E tests)

### Additional information

The root cause is that in java worker we are initialize some singleton in the first function load request, this customer has multiple functions in one function app, when there are multiple function load requests, for ex A & warmup trigger coming almost at same time, let's say A is the first request java worker take to process, before java worker start initialize the singleton object in function A load request (as it's the first load request that we set the atomic flag in it ), it switch to processing warmup trigger load request in a different thread and finished it and return to host, then host send function warmup trigger invocation while function load request A is still be processing by another thread in java worker and the single objects haven't been fully initialized yet. So the invocation of warmup trigger gives the null pointer exception.  
Warmup trigger function is load and invoked
MicrosoftTeams-image

before singleton object was initialized in first function load request 
MicrosoftTeams-image (1)

Why this issue is not happening before we adding the middleware feature. 

The reason is that before the middleware feature, we don't really care about which function load request finish first. Once there is one function load request finish, then we can start invocation request without any issue. 
But once we add middleware feature, we need the first function load request (because the first function load request will initialize the singleton to be used in invocation request in that onetime logic block) to finish before sending invocations. The first function load request is doing extra work by executing onetime logics to initialize singleton, if the onetime logics block is not synchronized, there is an chance that the first function load request won't be the first one to finish.

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
@@ -35,7 +35,7 @@ public class JavaFunctionBroker {
 	private final Map<String, ImmutablePair<String, FunctionDefinition>> methods;
 	private final ClassLoaderProvider classLoaderProvider;
 	private String workerDirectory;
-	private boolean oneTimeLogicInitialized = false;
+	private volatile boolean oneTimeLogicInitialized = false;
 	private volatile InvocationChainFactory invocationChainFactory;
 	private volatile FunctionInstanceInjector functionInstanceInjector;
 	private final Object lock = new Object();
@@ -64,11 +64,13 @@ public class JavaFunctionBroker {
 	}
 
 	private void initializeOneTimeLogics() {
-		synchronized (lock) {
-			if (!oneTimeLogicInitialized) {
-				initializeInvocationChainFactory();
-				initializeFunctionInstanceInjector();
-				oneTimeLogicInitialized = true;
+		if (!oneTimeLogicInitialized) {
+			synchronized (lock) {
+				if (!oneTimeLogicInitialized) {
+					initializeInvocationChainFactory();
+					initializeFunctionInstanceInjector();
+					oneTimeLogicInitialized = true;
+				}
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -63,9 +64,12 @@ public void loadMethod(FunctionMethodDescriptor descriptor, Map<String, BindingI
 	}
 
 	private void initializeOneTimeLogics() {
```
kaibocai, line 64: Hi @trask , I remember when I first implement this method I am using below code

```java
private void initializeOneTimeLogics() {
	if (!oneTimeLogicInitialized) {
		synchronized (lock) {
			if (!oneTimeLogicInitialized) {
				initializeInvocationChainFactory();
				initializeFunctionInstanceInjector();
				oneTimeLogicInitialized = true;
			}
		}
	}
}
```

I am using to if condition here (the idea is for performance improving), but this code is exact same as we use `AtomicBoolean`, it will still have this NPE issue, as we can skip the core initialize logics here. So with/without using `AtomicBoolean` we will get this issue anyway :)

trask: I don't think the above code would suffer the NPE, the second thread will block on the synchronized since you don't update the boolean until the end

kaibocai: I see, so as long as we don't release the lock in the synchronized block (which we don't), the all other threads have to be blocked until the first thread exit the synchronized block.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -35,9 +35,10 @@ public class JavaFunctionBroker {
 	private final Map<String, ImmutablePair<String, FunctionDefinition>> methods;
 	private final ClassLoaderProvider classLoaderProvider;
 	private String workerDirectory;
-	private final AtomicBoolean oneTimeLogicInitialized = new AtomicBoolean(false);
+	private volatile boolean oneTimeLogicInitialized = false;
 	private volatile InvocationChainFactory invocationChainFactory;
 	private volatile FunctionInstanceInjector functionInstanceInjector;
+	private final Object lock = new Object();
```
trask: i'd give this lock a name, e.g. `initializationLock`, to make it clear it shouldn't be used for any other locking purposes

trask (approved):

PR #692, opened 2023-01-07
Migrate from junit4 to junit5

kaibocai:

Migrate from junit4 to junit5

### Pull request checklist

* [X] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information


kamperiadis (dismissed):
LGTM but it looks like you have some build issues to resolve.

src/main/java/com/microsoft/azure/functions/worker/handler/InvocationRequestHandler.java
```diff
@@ -25,7 +25,6 @@ public InvocationRequestHandler(JavaFunctionBroker broker) {
 
     @Override
     String execute(InvocationRequest request, InvocationResponse.Builder response) throws Exception {
-        WorkerLogManager.getSystemLogger().log(Level.INFO, "Java version - " + getJavaVersion());
```
shreyas-gopalakrishna, line 28: Any reason this log is removed? This was added to make it easier to get java version in kusto logs specifically for consumption apps as they use placeholders. I suggest we retain this.

kaibocai: we can check the version by using log here https://github.com/Azure/azure-functions-java-worker/blob/86e4db1ce8870710290a5d17e11164a1dc568227/src/main/java/com/microsoft/azure/functions/worker/broker/JavaMethodExecutors.java#L9, and from applens we can also tell the version. 

My concern is that put it in invocation is not a good place, when cx is having thousands rps, this can be kind of annoying and it may impact the throughput and performance when the rps is high as we need to run getJavaverion very frequently and log them to host from stdout. 

So if we have a way to tell the java version, let's just save the resources as much as possible.

shreyas-gopalakrishna: Understand your concern. I feel it is better to retain it with a lower log level. This log specifically informs not only the java version but also the java path which is useful for the customer and us. Some customers had reached out asking how to get this information. The time we would save with one log is quite miniscule. Feel free to combine java version with the next log if it saves time.

kaibocai: I moved the log to function load request handler as function load is far fewer than function invocation.

shreyas-gopalakrishna (approved):

PR #694, opened 2023-01-20
Enable use of nullable maps for headers, params and query

### LINKED ISSUE #683: Empty string cannot be passed in HttpTrigger query string
#### Repro steps

1. Run the following function from templates:
```java
package com.function;

import java.util.*;
import com.microsoft.azure.functions.annotation.*;
import com.microsoft.azure.functions.*;

/**
 * Azure Functions with HTTP Trigger.
 */
public class HttpTriggerJava1 {
    /**
     * This function listens at endpoint "/api/HttpTriggerJava1". Two ways to invoke it using "curl" command in bash:
     * 1. curl -d "HTTP Body" {your host}/api/HttpTriggerJava1
     * 2. curl {your host}/api/HttpTriggerJava1?name=HTTP%20Query
     */
    @FunctionName("HttpTriggerJava1")
    public HttpResponseMessage run(
            @HttpTrigger(name = "req", methods = {HttpMethod.GET, HttpMethod.POST}, authLevel = AuthorizationLevel.FUNCTION) HttpRequestMessage<Optional<String>> request,
        context.getLogger().info("Java HTTP trigger processed a request.");

        // Parse query parameter
        String query = request.getQueryParameters().get("name");
        String name = request.getBody().orElse(query);

        if (name == null) {
            return request.createResponseBuilder(HttpStatus.BAD_REQUEST).body("Please pass a name on the query string or in the request body").build();
        } else {
            return request.createResponseBuilder(HttpStatus.OK).body("Hello, " + name).build();
        }
    }
}
```
2. Send the following request: [http://localhost:7071/api/HttpTriggerJava1?name=](http://localhost:7071/api/HttpTriggerJava1?name=)

#### Expected behavior
`name` should be set to an empty string and  the function should return "Hello, "

#### Actual behavior
`name` is set to null and the function returns "Please pass a name on the query string or in the request body"

#### Known workarounds
None

#### Related information 
Similar issue in PowerShell worker: https://github.com/Azure/azure-functions-powershell-worker/issues/895

kamperiadis:
BREAKING CHANGE: Enabled the `UseNullableValueDictionaryForHttp` capability which makes the host send back request information in nullable maps (i.e. When the `UseNullableValueDictionaryForHttp` capability is enabled, the host stores the values in `nullable_headers`, `nullable_params` and `nullable_query` instead of `headers`, `params` and `query`)

**Context:**
This is the sample function app code run for the screenshots below:
image
**Before:**
Without this capability, if the customer were to set an empty string as the query value, the host would not add the entry to the query map. This means that if the customer were to try to access that entry from the query map (via the `getQueryParameters` function), they would just get `null`.
beforeChanges
**After:**
With this capability, if the customer were to set an empty string as the query value, the host would add the entry to the query map with the value just being the empty string.
withChanges

This is meant to resolve issue #683 

### Pull request checklist

* [x] My changes **do not** need to be backported to a previous version
* [x] I have added all required tests (Unit tests, E2E tests)

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import com.microsoft.azure.functions.rpc.messages.NullableTypes;
+import com.microsoft.azure.functions.worker.Util;
 import org.apache.commons.lang3.reflect.TypeUtils;
 
 import com.microsoft.azure.functions.HttpMethod;
@@ -92,8 +93,14 @@ public final class RpcHttpRequestDataSource extends DataSource<RpcHttpRequestDat
 	}
 
 	private static Map<String, String> convertFromNullableMap(Map<String, NullableTypes.NullableString> nullableMap) {
-		return nullableMap.entrySet().stream().collect(
-			Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getValue())
-		);
+		if (Util.isTrue(System.getenv("FUNCTIONS_WORKER_NULLABLE_VALUES_ENABLED"))) {
+			return nullableMap.entrySet().stream().collect(
+					Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getValue())
+			);
+		} else {
+			return nullableMap.entrySet().stream()
+					.filter(e -> e.getValue().getValue() != null && !e.getValue().getValue().trim().isEmpty())
+					.collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getValue()));
+		}
 	}
 }
--- a/src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
@@ -6,7 +6,9 @@ import java.net.URI;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
+import com.microsoft.azure.functions.rpc.messages.NullableTypes;
 import org.apache.commons.lang3.reflect.TypeUtils;
 
 import com.microsoft.azure.functions.HttpMethod;
@@ -22,8 +24,10 @@ public final class RpcHttpRequestDataSource extends DataSource<RpcHttpRequestDat
 		super(name, null, HTTP_DATA_OPERATIONS);
 		this.httpPayload = value;
 		this.bodyDataSource = BindingDataStore.rpcSourceFromTypedData(null, this.httpPayload.getBody());
-		this.fields = Arrays.asList(this.httpPayload.getHeadersMap(), this.httpPayload.getQueryMap(),
-				this.httpPayload.getParamsMap());
+		this.fields = Arrays.asList(
+				convertFromNullableMap(this.httpPayload.getNullableHeadersMap()),
+				convertFromNullableMap(this.httpPayload.getNullableQueryMap()),
+				convertFromNullableMap(this.httpPayload.getNullableParamsMap()));
 		this.setValue(this);
 	}
 
@@ -45,12 +49,12 @@ public final class RpcHttpRequestDataSource extends DataSource<RpcHttpRequestDat
 
 		@Override
 		public Map<String, String> getHeaders() {
-			return this.parentDataSource.httpPayload.getHeadersMap();
+			return convertFromNullableMap(this.parentDataSource.httpPayload.getNullableHeadersMap());
 		}
 
 		@Override
 		public Map<String, String> getQueryParameters() {
-			return this.parentDataSource.httpPayload.getQueryMap();
+			return convertFromNullableMap(this.parentDataSource.httpPayload.getNullableQueryMap());
 		}
 
 		@Override
@@ -85,5 +89,11 @@ public final class RpcHttpRequestDataSource extends DataSource<RpcHttpRequestDat
 			BindingData bodyData = v.bodyDataSource.computeByType(actualType).orElseThrow(ClassCastException::new);
 			return new HttpRequestMessageImpl(v, bodyData.getValue());
 		});
+	}
+
+	private static Map<String, String> convertFromNullableMap(Map<String, NullableTypes.NullableString> nullableMap) {
+		return nullableMap.entrySet().stream().collect(
+			Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getValue())
+		);
 	}
 }
--- a/src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import com.microsoft.azure.functions.rpc.messages.NullableTypes;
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
```diff
@@ -22,8 +25,10 @@ public RpcHttpRequestDataSource(String name, RpcHttp value) {
 		super(name, null, HTTP_DATA_OPERATIONS);
 		this.httpPayload = value;
 		this.bodyDataSource = BindingDataStore.rpcSourceFromTypedData(null, this.httpPayload.getBody());
-		this.fields = Arrays.asList(this.httpPayload.getHeadersMap(), this.httpPayload.getQueryMap(),
-				this.httpPayload.getParamsMap());
+		this.fields = Arrays.asList(
+				convertFromNullableMap(this.httpPayload.getNullableHeadersMap()),
+				convertFromNullableMap(this.httpPayload.getNullableQueryMap()),
+				convertFromNullableMap(this.httpPayload.getNullableParamsMap()));
```
kaibocai, line 32: not related to this PR, but do we know difference between query map and param map?

src/test/java/com/microsoft/azure/functions/worker/binding/tests/RpcHttpRequestDataSourceTest.java
```diff
@@ -41,14 +56,150 @@ public static RpcHttp getTestRpcHttp(Object inputBody) throws Exception {
     return httpBuilder.build();
   }
 
+  @Test
+  public void rpcHttpDataSource_To_HttpRequestMessage_NullableQueryParamsEmpty_EnvSettingEnabled() throws Exception {
```
kaibocai: let's use camel-case syntax for method name, the old tests were wrote by c# expert, seems they are using c# name convention.

src/test/java/com/microsoft/azure/functions/worker/binding/tests/RpcHttpRequestDataSourceTest.java
```diff
@@ -41,14 +56,150 @@ public static RpcHttp getTestRpcHttp(Object inputBody) throws Exception {
...
+    }};
+    Parameter[] parameters = httpRequestMessageStringBodyMethod.getParameters();
+    String sourceKey = "testRpcHttp";
+    RpcHttp input = getTestRpcHttp(null, null, queryMap);
+    RpcHttpRequestDataSource rpcHttp = new RpcHttpRequestDataSource(sourceKey, input);
+    Optional<BindingData> actualBindingData = rpcHttp.computeByName(sourceKey,
+            parameters[0].getParameterizedType());
+    BindingData actualArg = actualBindingData.orElseThrow(WrongMethodTypeException::new);
+    HttpRequestMessage<?> requestMsg = (HttpRequestMessage<?>) actualArg.getValue();
+    assertEquals(requestMsg.getQueryParameters().get("name"), "");
```
kaibocai: let's follow the convention that expect value comes first and test value comes second. I think we can use `assertEmpty` here.

src/test/java/com/microsoft/azure/functions/worker/binding/tests/RpcHttpRequestDataSourceTest.java
```diff
@@ -41,14 +56,150 @@ public static RpcHttp getTestRpcHttp(Object inputBody) throws Exception {
...
+    }};
+    Parameter[] parameters = httpRequestMessageStringBodyMethod.getParameters();
+    String sourceKey = "testRpcHttp";
+    RpcHttp input = getTestRpcHttp(null, null, queryMap);
+    RpcHttpRequestDataSource rpcHttp = new RpcHttpRequestDataSource(sourceKey, input);
+    Optional<BindingData> actualBindingData = rpcHttp.computeByName(sourceKey,
+            parameters[0].getParameterizedType());
+    BindingData actualArg = actualBindingData.orElseThrow(WrongMethodTypeException::new);
+    HttpRequestMessage<?> requestMsg = (HttpRequestMessage<?>) actualArg.getValue();
+    assertEquals(requestMsg.getQueryParameters().get("name"), null);
```
kaibocai: we can use assertNull

src/test/java/com/microsoft/azure/functions/worker/binding/tests/RpcHttpRequestDataSourceTest.java
```diff
@@ -41,14 +56,150 @@ public static RpcHttp getTestRpcHttp(Object inputBody) throws Exception {
...
+  }
+
+  @Test
+  public void rpcHttpDataSource_To_HttpRequestMessage_NullableQueryParamsNonEmpty() throws Exception {
```
kaibocai: I think we can have the test cases more clear. There are two major scenarios we want to test
1. appsetting to false
2. appsettting to true

each of those scenarios we want to test
1. normal case -- param value empty
2. param value empty

So we can have two test cases in total
1. with appsetting to true

- have one normal param -- expect same value
- have one param with empty value -- expect empty string
2. with appsetting to false

- have one normal param -- expect normal value
- have one param with empty value -- expect null

We can test header map and query map in same test method. 
So we can reduce the test methods number to 2 but still have full coverage. Let me know if this make sense or not. Thanks.

src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
kamperiadis: Looking at the host code, it looks like the three key differences between [query map](https://github.com/Azure/azure-functions-host/blob/11a0add2eb0c14588e5070aed3d0489782762be8/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L116-L130) and [param map](https://github.com/Azure/azure-functions-host/blob/11a0add2eb0c14588e5070aed3d0489782762be8/src/WebJobs.Script.Grpc/MessageExtensions/GrpcMessageConversionExtensions.cs#L149-L166) are that:
1. They come from different parts of the `request` object (i.e. `request.Query` vs `request.HttpContext.Items.TryGetValue(HttpExtensionConstants.AzureWebJobsHttpRouteDataKey, out object routeData)`)
2. We do not add null values to the parameters map, even if the `UseNullableValueDictionaryForHttp` capability is on 
3. We add empty strings to the parameters map, even if the `UseNullableValueDictionaryForHttp` capability is off

kaibocai (dismissed):

kaibocai:
Once this feature is released, need to document it somewhere (repo wiki?) for customer. Maybe create an issue to track creating the document.

src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
```diff
@@ -86,4 +91,16 @@ public Builder createResponseBuilder(HttpStatus status) {
 			return new HttpRequestMessageImpl(v, bodyData.getValue());
 		});
 	}
+
+	private static Map<String, String> convertFromNullableMap(Map<String, NullableTypes.NullableString> nullableMap) {
+		if (Util.isTrue(System.getenv("FUNCTIONS_WORKER_NULLABLE_VALUES_ENABLED"))) {
```
shreyas-gopalakrishna: Let us make `FUNCTIONS_WORKER_NULLABLE_VALUES_ENABLED` a constant.

shreyas-gopalakrishna (dismissed):
LGTM! Just one minor refactor

src/main/java/com/microsoft/azure/functions/worker/binding/RpcHttpRequestDataSource.java
kaibocai: nice catch! there is a constant file you can define it there.

Issue #682, opened 2022-11-25
Feature Request/Improvement - Add Gson type adapter for java.time.Instant


ffarinha-msft:
When using the feature that some functions triggers implement, of allowing to received data as POJO ([like Event Grid trigger ](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-grid-trigger?tabs=in-process%2Cextensionv3&pivots=programming-language-java#event-grid-trigger-pojo-parameter) ) we found that we cannot deserialize the JSON payload to the type java.time.Instant and if you try you will get the below error:

`InaccessibleObjectException: Unable to make field private final long java.time.Instant.seconds accessible: module java.base does not "opens java.time" to unnamed module @41b0ae4c Stack: com.google.gson.JsonIOException: Failed making field 'java.time.Instant#seconds' accessible; either change its visibility or write a custom TypeAdapter for its declaring type at com.google.gson.internal.reflect.ReflectionHelper.makeAccessible(ReflectionHelper.java:22) at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:158) at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:101) at com.google.gson.Gson.getAdapter(Gson.java:501) at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.createBoundField(ReflectiveTypeAdapterFactory.java:116) at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:165) at`

This is because Function Java Worker uses GSON library to deserialize the JSON input and this library has historically not been able to deserialize the data to the type java.time.Instant, the Function Java Worker Runtime also inherited this limitation.

Relevant articles:
•	https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-java?tabs=bash%2Cconsumption#pojos
•	[GSON Serialization/Deserialization of Java 8 Date API · Issue #1059 · google/gson (github.com)](https://github.com/google/gson/issues/1059)

The java.time.Instant has been released quite a while ago and it should replace the usage of java.util.Date.

#### Repro steps

[...]

Issue #419, opened 2021-02-25
gson response date-format


ODH003:
Hi team, I'm using azure cloud function java 11 and the response of the functions are ok, just I would like to change the date-format for all output object. But I couldn't find any information about it.

![image](https://user-images.githubusercontent.com/48140216/109092639-ea45dc00-76e4-11eb-8715-247c21db8168.png)

I would like to get the response format:
spring.gson.date-format=yyyy-MM-dd'T'HH:mm:ss

Like left json of the picture.

CrawX:
@ODH003 fyi I posted a workaround that could maybe help you out in the feature request #424.

kaibocai:
Hi @ODH003 , Thanks very much for opening this issue. As CrawX provided a workaround, we still want to know if you can provide an repo used to reproduce this issue. Thanks very much.

PR #704, opened 2023-02-16
potential support for customize gson

### LINKED ISSUE #424: [FeatureRequest] Allow customizing the Gson instance for Json serialization
Hi all,

**Current problem**
The Gson instance used by `RpcUnspecifiedDataTarget` is hardcoded in `RpcJsonDataSource` 
https://github.com/Azure/azure-functions-java-worker/blob/3ec27b2b82ba96adef593d51b0bb0e9c0063bfaa/src/main/java/com/microsoft/azure/functions/worker/binding/RpcUnspecifiedDataTarget.java#L68
https://github.com/Azure/azure-functions-java-worker/blob/3ec27b2b82ba96adef593d51b0bb0e9c0063bfaa/src/main/java/com/microsoft/azure/functions/worker/binding/RpcJsonDataSource.java#L20
This means that customizing the way the runtime Json-serializes objects (especially dates) is impossible. See #419 and [this on stackoverflow](https://stackoverflow.com/questions/60688436/spring-cloud-function-azure-json-output-with-datetimes).

**Workaround**
I came up with a dirty workaround to replace the used Gson instance with the spring-provided one via Reflection:

`AzureGsonConfig.java`:
```java
@Configuration
@ConditionalOnClass(RpcJsonDataSource.class)
public class AzureGsonConfig {

  @SneakyThrows
  public AzureGsonConfig(Gson gson) {
    final Field gsonField = RpcJsonDataSource.class.getField("gson");
    gsonField.setAccessible(true);
    final Field modifiersField = Field.class.getDeclaredField("modifiers");
    modifiersField.setAccessible(true);
    modifiersField.setInt(gsonField, gsonField.getModifiers() & ~Modifier.FINAL);

    gsonField.set(null, gson);
  }
}
```

This requires a compile-time dependency like this
```grovy
compileOnly files('libs/azure-functions-java-worker.jar')
```
(referencing the runtime as a jar file because I couldn't find it in maven central).

Having replaced the Gson instance means you can have a second configuration that customizes the date serialization behaviour of the one from spring.

`GsonConfig.java`:
```java
@Configuration
public class GsonConfig {

  @Bean
  GsonBuilderCustomizer dateFormatGsonBuilderCustomizer() {
    return gsonBuilder -> {
      gsonBuilder.registerTypeAdapter(
          LocalDate.class, new LocalDateAdapter()
      );
      gsonBuilder.registerTypeAdapter(
          LocalDateTime.class, new LocalDateTimeAdapter()
      );
    };
  }

  static class LocalDateAdapter extends TypeAdapter<LocalDate> {

    @Override
    public void write(final JsonWriter jsonWriter, final LocalDate localDate) throws IOException {
      if (localDate == null) {
        jsonWriter.nullValue();
      } else {
        jsonWriter.value(localDate.toString());
      }
    }

    @Override
    public LocalDate read(final JsonReader jsonReader) throws IOException {
      if (jsonReader.peek() == JsonToken.NULL) {
        jsonReader.nextNull();
        return null;
      } else {
        return LocalDate.parse(jsonReader.nextString());
      }
    }
  }

  static class LocalDateTimeAdapter extends TypeAdapter<LocalDateTime> {

    @Override
    public void write(final JsonWriter jsonWriter, final LocalDateTime localDateTime) throws IOException {
      if (localDateTime == null) {
        jsonWriter.nullValue();
      } else {
        jsonWriter.value(localDateTime.toString());
      }
    }

    @Override
    public LocalDateTime read(final JsonReader jsonReader) throws IOException {
      if (jsonReader.peek() == JsonToken.NULL) {
        jsonReader.nextNull();
        return null;
      } else {
        return LocalDateTime.parse(jsonReader.nextString());
      }
    }
  }
}
```

While this works, it's not very clean and can break easily.

**Alternative**
Serialize the objects to strings in the functions or handlers and return them as such. Personally, I don't like this approach.

**Feature request**
Somehow allow injecting or customizing the Gson instance used in the azure-functions-worker runtime.

kaibocai:

A primitive implementation for support customizing gson instance used in java worker. 
Issues opened on hard code gson instance are many: https://github.com/Azure/azure-functions-java-worker/issues/682, https://github.com/Azure/azure-functions-java-worker/issues/424, https://github.com/Azure/azure-functions-java-worker/issues/419, https://github.com/Azure/azure-functions-host/issues/7391, etc.

So maybe it's good to add flexibility supporting customer customize their gson instance. 
Another solution would be work with Azure java sdk team on moving out of gson with better choice. 
Board link: https://msazure.visualstudio.com/Antares/_boards/board/t/Functions%20Java/Backlog%20items/?workitem=17209424

SPI requirement: https://github.com/Azure/azure-functions-java-additions/pull/20

**This feature may not be needed at all, as most customer can using plan String and customize their gson instance inside their function to deserialize.** 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/Util.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/Util.java
@@ -1,6 +1,11 @@
 package com.microsoft.azure.functions.worker;
 
+import com.google.gson.Gson;
+import com.microsoft.azure.functions.spi.inject.GsonInstanceInjector;
+
 public class Util {
+    private static Gson gsonInstance;
+    private static final Object utilLock = new Object();
     public static boolean isTrue(String value) {
         if(value != null && (value.equalsIgnoreCase("true") || value.equalsIgnoreCase("1"))) {
             return true;
@@ -11,4 +16,19 @@ public class Util {
     public static String getJavaVersion() {
         return String.join(" - ", System.getProperty("java.home"), System.getProperty("java.version"));
     }
+
+    public static void setGsonInstance(Gson instance) {
+        if (gsonInstance == null) {
+            synchronized (utilLock) {
+                if (gsonInstance == null) {
+                    gsonInstance = instance;
+                }
+            }
+        }
+    }
+
+    public static Gson getGsonInstance() {
+        return gsonInstance;
+    }
+
 }
--- a/src/main/java/com/microsoft/azure/functions/worker/Util.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/Util.java
@@ -18,13 +18,7 @@ public class Util {
     }
 
     public static void setGsonInstance(Gson instance) {
-        if (gsonInstance == null) {
-            synchronized (utilLock) {
-                if (gsonInstance == null) {
-                    gsonInstance = instance;
-                }
-            }
-        }
+        gsonInstance = instance;
     }
 
     public static Gson getGsonInstance() {
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <azure.functions.java.core.library.version>1.2.0</azure.functions.java.core.library.version>
-    <azure.functions.java.spi>1.0.0</azure.functions.java.spi>
+    <azure.functions.java.spi>1.1.0</azure.functions.java.spi>
     <azure.functions.java.library.version>2.2.0</azure.functions.java.library.version>
     <jupiter.version>5.9.1</jupiter.version>
     <mockito-core.version>4.11.0</mockito-core.version>
--- a/src/main/java/com/microsoft/azure/functions/worker/binding/RpcJsonDataSource.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/binding/RpcJsonDataSource.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import com.microsoft.azure.functions.worker.Util;
 import org.apache.commons.lang3.reflect.TypeUtils;
 
 import com.google.gson.Gson;
@@ -17,7 +18,7 @@ public final class RpcJsonDataSource extends DataSource<String> {
 		super(name, value, JSON_DATA_OPERATIONS);
 	}
 
-	public static final Gson gson = new Gson();
+	public static final Gson gson = Util.getGsonInstance();
 	public static final JsonParser gsonParser = new JsonParser();
 	private static final DataOperations<String, Object> JSON_DATA_OPERATIONS = new DataOperations<>();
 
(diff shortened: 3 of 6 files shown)


liliankasem:
> This feature may not be needed at all, as most customer can using plan String and customize their gson instance inside their function to deserialize.

Customers can just bind to string and do their own deserialization, but I feel that this breaks the POJO scenario. When binding to a POJO, we should be able to handle that properly or allow for customers to customize gson as suggested here. This will avoid exceptions when deserializing something like `Date` 

```log
[2023-02-16T00:31:00.253Z] Exception: IllegalArgumentException: No time zone indicator
[2023-02-16T00:31:00.253Z] Stack: com.google.gson.JsonSyntaxException: Failed parsing '0001-01-01T00:00:00' as Date; at path $.ScheduleStatus
```

I think we should avoid guiding customers to use string binding when we can't deserialize their payload because of something like a date format, that's more an issue with us than with their code.

liliankasem:
> > This feature may not be needed at all, as most customer can using plan String and customize their gson instance inside their function to deserialize.
> 
> Customers can just bind to string and do their own deserialization, but I feel that this breaks the POJO scenario. When binding to a POJO, we should be able to handle that properly or allow for customers to customize gson as suggested here. This will avoid exceptions when deserializing something like `Date`
> 
> ```
> [2023-02-16T00:31:00.253Z] Exception: IllegalArgumentException: No time zone indicator
> [2023-02-16T00:31:00.253Z] Stack: com.google.gson.JsonSyntaxException: Failed parsing '0001-01-01T00:00:00' as Date; at path $.ScheduleStatus
> ```
> 
> I think we should avoid guiding customers to use string binding when we can't deserialize their payload because of something like a date format, that's more an issue with us than with their code.

@kaibocai The PR to fix the default DateTime for the timer info has been merged and should be released within the next month or so. However, I would still recommend this support for custom gson builder to avoid issues like this with other types and to provide user's more flexibility.

kaibocai:
> @kaibocai The PR to fix the default DateTime for the timer info has been merged and should be released within the next month or so. However, I would still recommend this support for custom gson builder to avoid issues like this with other types and to provide user's more flexibility.

yes, we definitely need to provide flexibility to customize serializing and deserializing.

src/main/java/com/microsoft/azure/functions/worker/Util.java
```diff
@@ -11,4 +16,19 @@ public static boolean isTrue(String value) {
     public static String getJavaVersion() {
         return String.join(" - ", System.getProperty("java.home"), System.getProperty("java.version"));
     }
+
+    public static void setGsonInstance(Gson instance) {
+        if (gsonInstance == null) {
+            synchronized (utilLock) {
```
kaibocai: No need to add sync block here, it's already thread safe with the sync block at https://github.com/Azure/azure-functions-java-worker/blob/67258975a86bca1d7a356cad8c361ab79551a179/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java#L69

shreyas-gopalakrishna:
Issues - https://github.com/Azure/azure-functions-java-worker/issues/720

PR #790, opened 2024-10-08
emulated tests support

YunchuWang:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information


gavin-aguiar (approved):

eng/ci/public-build.yml
```diff
@@ -42,4 +42,16 @@ extends:
...
+      - template: /eng/ci/templates/jobs/build.yml@self
+    - stage: TestWindows
+      dependsOn: Build
+
+      jobs:
+      - template: /eng/ci/templates/jobs/run-emulated-tests-windows.yml@self
+
+    - stage: TestLinux
+      dependsOn:
+      - TestWindows
```
hallvictoria, line 54: We should run the windows and linux emulators tests in parallel instead of sequential

hallvictoria: Also should we run the emulators test as part of the official build as well?

emulatedtests/utils/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
+name: microsoft-azure-eventhubs
+services:
+  # Service for the Event Hubs Emulator
+  emulator:
+    container_name: "eventhubs-emulator"
```
hallvictoria, line 5: Is there a reason why EventHub tests aren't being run? If so, we can remove the EventHub related setup from this file

emulatedtests/utils/Config.json
```diff
@@ -0,0 +1,51 @@
+{
+    "UserConfig": {
+        "NamespaceConfig": [
```
hallvictoria, line 3: The EventHub names would most likely need to be changed for java tests, but if EventHub isn't being tested we don't need this file

eng/ci/public-build.yml
YunchuWang: unfortunately the tests are executed against same set of resources. i dont have previous context but i guess it could bring race condition issues. its something can be tried later.
for official build i think its better to use real resource for internal e2e test for full verification. this is for public more 
image

emulatedtests/utils/Config.json
YunchuWang: could be used for next pr to add eventhub

emulatedtests/utils/docker-compose.yml
YunchuWang: could be used for next pr to add eventhub

PR #838, opened 2025-10-03
Use official Microsoft build of JDK 25

ahmedmuhsin:
This pull request updates the JDK 25 configuration for both Linux and Windows emulated test jobs in the CI pipeline and refactors the way JDK versions are managed in the CI pipeline templates for both Linux and Windows. The main improvement is the centralization of JDK version variables into a separate YAML file, which is then imported into the job templates. This change makes it easier to update JDK versions in the future and ensures consistency across the pipeline. Additionally, the JDK 25 test matrix entries have been updated to use Microsoft builds instead of Adoptium builds.

Key changes include:

**JDK 25 configuration updates**
* Replaced Adoptium JDK 25 early access references with Microsoft JDK 25.0.0 in the Linux emulated test job, updating the download link and JAVA_VERSION name accordingly.
* Replaced Adoptium JDK 25 early access references with Microsoft JDK 25.0.0 in the Windows emulated test job, updating the download link and JAVA_VERSION name accordingly.

**Centralization and Reuse:**

* Created a new `java-versions.yml` file that defines all JDK version variables for both Linux and Windows. These variables are now referenced throughout the CI templates, improving maintainability and consistency.
* Updated both `run-emulated-tests-linux.yml` and `run-emulated-tests-windows.yml` to import the centralized JDK version variables using the new template. [[1]](diffhunk://#diff-bda3e01807ffd61592003b12ef34cd8265b418dbd808b93884135b8b81f91e76L16-R44) [[2]](diffhunk://#diff-28a39034bb284340b17656330e08b33af3cce0ee604d4d0f9d238d7315e75d68L16-R44)

**Test Matrix Updates:**

* Modified all JDK download links and version identifiers in the test matrices to use the centralized variables, reducing duplication and risk of inconsistency. [[1]](diffhunk://#diff-bda3e01807ffd61592003b12ef34cd8265b418dbd808b93884135b8b81f91e76L16-R44) [[2]](diffhunk://#diff-28a39034bb284340b17656330e08b33af3cce0ee604d4d0f9d238d7315e75d68L16-R44)
* Replaced Adoptium JDK 25 entries with Microsoft JDK 25 builds in both Linux and Windows test matrices, standardizing the JDK provider for version 25. [[1]](diffhunk://#diff-bda3e01807ffd61592003b12ef34cd8265b418dbd808b93884135b8b81f91e76L16-R44) [[2]](diffhunk://#diff-28a39034bb284340b17656330e08b33af3cce0ee604d4d0f9d238d7315e75d68L16-R44)


gavin-aguiar (dismissed):

eng/ci/templates/jobs/run-emulated-tests-linux.yml
```diff
@@ -13,33 +13,35 @@ jobs:
       os: linux
 
     variables:
-      ${{ if contains(variables['Build.SourceBranch'], '/tags/' ) }}:
-        isTagTemp: true
-      isTag: $[variables.isTagTemp]
-      ApplicationInsightAgentVersion: 3.5.4
+      - template: ../java-versions.yml
```
TsuyoshiUshio, line 16: Not a request. but question.

Why did you make this change? I'd like to understand the context. 
* Why did you change the logic of isTagTemp assignment? The previous logic has a problem?
* Why did you remove ApplicationInsightsAgentVersion? Not used before?

TsuyoshiUshio (dismissed):

eng/ci/templates/jobs/run-emulated-tests-linux.yml
ahmedmuhsin: The logic for isTag and isTagTemp has not changed. I did however have to change the syntax a little after adding the java-version.yml template. Instead of the shorthand that was previously there now I have to specify name and value for each variable. 

I added java-versions.yml to manage all of the JDK versions in one place so I can later add a pipeline to automatically check for new JDK build versions and update the file. That way, the pipeline just has to update one file which both of the test templates reference.

TsuyoshiUshio (approved):

PR #841, opened 2025-11-21
Fix specialization bugs

ahmedmuhsin:
This pull request ensures that the Otel capabilities are updated during specialization and the SDK types feature flag is correctly processed. It also introduces a new Java-based Azure Functions sample app with an Event Grid Blob trigger, along with supporting scripts and configuration for Docker-based integration testing. The changes enable streamlined building, packaging, and testing of Azure Functions Java apps in containerized environments, including support for SDK type bindings and local Azurite storage emulation.

**New Java Azure Functions Sample App:**
* Added a complete sample Azure Functions app (`BlobSdkType`) using Java with an Event Grid Blob trigger and SDK type bindings. This includes `pom.xml` for Maven configuration, `host.json` for extension bundles, and the main function logic in `ProcessBlobUpload.java`. The function listens for new blobs, processes them, and copies them to a processed container. [[1]](diffhunk://#diff-0d25417512f2420aeeddf3f269fedaa7f4582a8cc39552529f6ffb00745c5985R1-R102) [[2]](diffhunk://#diff-179709796aefe53fb9bbce919be59fba78cdcf5a7396f2e3b7ee8e263a9c10d6R1-R7) [[3]](diffhunk://#diff-2c0df82d7c3887a1e85d547c8d71ee2ccef9b920ef1564de221e4537817a728eR1-R64)

**Build and Packaging Automation:**
* Introduced `build-apps.ps1` PowerShell script to automate building Java apps and packaging them into squashfs images using Docker, improving reproducibility and simplifying local development.

**Testing Infrastructure and Configuration:**
* Added `.env.example` for environment variable configuration, supporting runtime selection, worker/app directories, and storage options for test environments.
* Added `pyproject.toml` to specify Python dependencies for integration tests, including Azure SDK, pytest, and other utilities.
* Provided `example_usage.py` demonstrating how to launch containers, assign packages, and verify function behavior and logs in a test environment.


src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -47,10 +47,9 @@ public class JavaFunctionBroker {
 	private List<Middleware> serviceLoadedMiddlewares = new ArrayList<>();
 	private final Map<String, InvocationChainFactory> functionFactories = new ConcurrentHashMap<>();
 	private final SdkParameterAnalyzer sdkParameterAnalyzer = new SdkParameterAnalyzer();
-	private final WorkerObjectCache<CacheKey> workerObjectCache;
-	private static final boolean JAVA_ENABLE_SDK_TYPES_FLAG =
```
swapnil-nagar, line 51: What was wrong with this? At runtime value is not fixed and how  pushing to synchronized block solves this?

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -320,4 +315,22 @@ private boolean isTesting(){
...
 		this.workerDirectory = workerDirectory;
 	}
+
+	private boolean isJavaSdkTypesEnabled() {
+		if (cachedSdkTypesEnabled == null) {
+			synchronized (this) {
+				if (cachedSdkTypesEnabled == null) {
+					String value = System.getenv(JAVA_ENABLE_SDK_TYPES);
+					// default to true
+					cachedSdkTypesEnabled = value == null ? true : Boolean.parseBoolean(value);
```
swapnil-nagar: Note: if customer has a garbage value other than "true" will result in cachedSdkTypesEnabled to false. Is that intended?

example Boolean.parseBoolean("DummmyValue") = false.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
```diff
@@ -320,4 +315,22 @@ private boolean isTesting(){
...
+					WorkerLogManager.getSystemLogger().info(
+						"Initialized SDK types enabled flag: " + cachedSdkTypesEnabled + 
+						" (from '" + JAVA_ENABLE_SDK_TYPES + "' environment variable : " + (value != null ? "'" + value + "'" : "null") + ")"
```
swapnil-nagar, line 328: [NIT]: Do you need to log value in the log? Your logic as cachedSdkTypesEnabled already determine that value, isn't need?

src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
```diff
@@ -31,6 +35,17 @@ String execute(FunctionEnvironmentReloadRequest request, Builder response) throw
 			return "Ignoring FunctionEnvironmentReloadRequest as newSettings map is empty.";
 		}
 		setEnv(environmentVariables);
+		
+		// Check for telemetry capabilities from environment variables
+		String openTelemetryEnabled = environmentVariables.get(JAVA_ENABLE_OPENTELEMETRY);
```
swapnil-nagar, line 47: Lets move OTEL variables to line 79 or below or create a separate method to evaluate those settings and referenced at line 80.

dockertests/app-src/BlobSdkType/host.json
```diff
@@ -0,0 +1,7 @@
+{
+  "version": "2.0",
+  "extensionBundle": {
+    "id": "Microsoft.Azure.Functions.ExtensionBundle",
+    "version": "[4.26.2]"
```
hallvictoria: Do we want to pin the specific extension bundle version? Might be worth considering if we can make that configurable as well - something that can be set in the .env file

dockertests/example_usage.py
```diff
@@ -0,0 +1,129 @@
...
+                'JAVA_ENABLE_OPENTELEMETRY': 'true',
+                'AzureWebJobsStorage': env.docker_storage_connection_string,
+                'PDFProcessorSTORAGE': env.docker_storage_connection_string
+            }
+        )
+        
+        # Wait for functions to be loaded
+        if env.functions_controller.wait_for_host_running(timeout=120):
+            # Functions are loaded, proceed with tests
+            test_sdk_types_flag_in_logs(env.functions_controller)
```
hallvictoria: Checking - `test_sdk_types_flag_in_logs` would fail here since `JAVA_ENABLE_SDK_TYPES` was set to false on line 44? Should we have this test commented out instead of `test_function_endpoint`?

dockertests/utils/azurite_container_controller.py
```diff
@@ -0,0 +1,182 @@
...
+import time
+from typing import Optional
+
+
+class AzuriteContainerController:
+    """Controller for managing Azurite storage emulator container."""
+    
+    # Fixed account key for Azurite
+    AZURITE_ACCOUNT_KEY = "Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=="
+
```
hallvictoria: nit: could this be a pipeline variable instead? I'm worried that it would get flagged by CodeQL, even if it's an emulator value

dockertests/utils/functions_container_controller.py
```diff
@@ -0,0 +1,615 @@
...
+from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import padding
+
+# Docker constants
+_DOCKER_PATH = "DOCKER_PATH"
+_DOCKER_DEFAULT_PATH = "docker"
+_MESH_IMAGE_URL = "https://mcr.microsoft.com/v2/azure-functions/mesh/tags/list"
+_MESH_IMAGE_REPO = "mcr.microsoft.com/azure-functions/mesh"
```
hallvictoria, line 28: future change: this can be changed to use flex instead of consumption

dockertests/utils/functions_container_controller.py
```diff
@@ -0,0 +1,615 @@
...
+        Args:
+            env: Environment variables to set in the container
+        """
+        # Add compulsory fields in specialization context
+        env["FUNCTIONS_EXTENSION_VERSION"] = f"~{self._host_version}"
+        env["FUNCTIONS_WORKER_RUNTIME"] = self._runtime
+        env["FUNCTIONS_WORKER_RUNTIME_VERSION"] = self._runtime_version
+        env["WEBSITE_SITE_NAME"] = self._site_name
+        env["WEBSITE_HOSTNAME"] = f"{self._site_name}.azurewebsites.com"
+        env["AzureWebEncryptionKey"] = "0F75CA46E7EBDD39E4CA6B074D1F9A5972B849A55F91A248"
```
hallvictoria: Please make this a pipeline variable

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
ahmedmuhsin: For Consumption SKUs the value is not passed in until the worker is specialized however, the JavaFunctionBroker instance is created before specialization.

The first time the value is needed is when a FunctionLoadRequest is received after specialization has occurred. The reason why we need a synchronized block is because the Host sends multiple FunctionLoadRequest at the same time, one per Function in the Function App but we only need to read the value once and cache it instead of reading it multiple times.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
ahmedmuhsin: Yes, that is intended. The value of the flag should not be interpreted as false unless they explicitly set it to 'false'. Otherwise, it will be true.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
ahmedmuhsin: I added this since it will be helpful for debugging especially while the feature is in Preview. We have faced a hard time debugging due to the lack of a log message like this.

ahmedmuhsin: This message will only appear once per app restart.

dockertests/utils/functions_container_controller.py
ahmedmuhsin: I updated the code so that the key is generated dynamically.

dockertests/utils/azurite_container_controller.py
ahmedmuhsin: I updated the code so that the key is generated dynamically.

dockertests/example_usage.py
ahmedmuhsin: I removed the test from the example.

dockertests/app-src/BlobSdkType/host.json
ahmedmuhsin: Fixed.

src/main/java/com/microsoft/azure/functions/worker/handler/FunctionEnvironmentReloadRequestHandler.java
ahmedmuhsin: I've refactored the logic into its own method.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
swapnil-nagar: But this logic is opposite of what you intended to achieve. 

For all the dummyValues which is not true will result in setting to false.

so if value is "DummyValue" or "xyz"

cachedSdkTypesEnabled will be false.

ahmedmuhsin: Ah, I see your point now. I'll fix it.

ahmedmuhsin: Fixed the issue and added it to the test cases.

src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
swapnil-nagar: So shouldn't you need invoke at the load time, that is
volatile Boolean cachedSdkTypesEnabled = isJavaSdkTypesEnabled() or during the constructor definition?

ahmedmuhsin: I need to defer getting that variable value until the first loadMethod() call since it is at that point that the worker is guaranteed to be specialized and has the correct value for the flag.

swapnil-nagar (approved):

PR #872, opened 2026-05-12
Honor secure Functions host URIs in Java worker

larohra:
## Summary
`JavaWorkerClient` always builds `ManagedChannelBuilder.forAddress(...).usePlaintext()`, so the Java worker ignores the already-parsed `--functions-uri` and can never join a secure host/worker gRPC channel.

## Findings
- **Fix necessary:** Yes. Today the worker guarantees plaintext and blocks any host-side TLS rollout, leaving any non-local transport exposed to MITM tampering.
- **Historical context:** Plaintext dates back to the early 2018 same-machine child-process design. In 2023 the worker added prefixed startup args, including `functions-uri`, plus fallback to legacy args, but `JavaWorkerClient` was not updated; this looks like an old assumption that became a gap once URI-based startup existed.
- **TLS behavior clarification:** The fix should honor TLS, not ignore it. When `functions-uri` uses `https`, the worker should build a TLS gRPC channel and fail if handshake, certificate, or hostname validation fails. Only `http` URIs, or legacy startup that supplies just host+port, should continue to use plaintext.
- **Regression / breaking change:** No new CLI or protocol contract. Existing `http` and legacy host+port launches keep their current behavior. The only observable behavior change is that a previously ignored or misconfigured `https` endpoint will stop connecting insecurely and instead fail closed.
- **Customer contract:** No new flags are required. Trust still comes from the JVM trust configuration already available via existing Java options, so hosts using private CAs do not need a new worker-specific switch.
- **Testing today:** `mvn test` currently passes (63 tests, 0 failures/errors/skips). Repo CI also runs build plus emulated, docker, and end-to-end matrices, but there is no direct secure gRPC transport coverage.

## Plan
1. Expose the parsed `functions-uri` through `IApplication` in a compatibility-safe way.
2. Make `JavaWorkerClient` choose transport from the URI scheme: `https` => TLS with no plaintext downgrade, `http` or legacy host+port => plaintext.
3. Add focused plaintext/TLS transport tests and rerun `mvn test`.

resolves #issue_for_this_pr

### Pull request checklist

* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information


src/test/java/com/microsoft/azure/functions/worker/functional/tests/GrpcTransportTest.java
```diff
@@ -0,0 +1,118 @@
...
+    public String ReturnStringFunction() {
+        return RETURN_VALUE;
+    }
+
+    @Test
+    public void legacyPlaintextTransportStillWorks() throws Exception {
+        System.setProperty("azure.functions.worker.java.skip.testing", "true");
+        try (FunctionsTestHost host = new FunctionsTestHost()) {
+            InvocationResponse response = this.invokeReturnString(host, "plaintext-function", "plaintext-request");
+
```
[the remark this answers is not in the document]
larohra: @copilot apply changes based on [this feedback](https://github.com/Azure/azure-functions-java-worker/pull/872#discussion_r3228705589)

src/main/java/com/microsoft/azure/functions/worker/JavaWorkerClient.java
```diff
@@ -138,4 +145,39 @@ private synchronized void send(String requestId, MessageHandler<?, ?> marshaller
...
+    /**
+     * @param functionsUri Host endpoint URI, or null for legacy startup args that only provide host and port.
+     */
+    static boolean useTransportSecurity(String functionsUri) {
+        if (functionsUri == null) {
+            return false;
+        }
+
+        String scheme = parseFunctionsUriScheme(functionsUri);
+        switch (scheme.toLowerCase(Locale.ROOT)) {
```
ahmedmuhsin: Consolidate URI parsing into `Application.parseUri` as the single source of truth. Validate the scheme there, default the port from the scheme, and decompose host/port. `JavaWorkerClient.useTransportSecurity` then collapses to a one-line scheme check. Today there are two parsers (`URL` in `Application`, `URI` in `JavaWorkerClient`) with different acceptance rules, and `https://host/` without an explicit port is rejected by `URL.getPort() == -1`. Both go away with the consolidation. Small change, fits naturally with the rest of this PR.

src/main/java/com/microsoft/azure/functions/worker/Application.java
```diff
@@ -157,10 +157,15 @@ public int getPort() {
         return this.port;
     }
 
-    public String getUri() {
+    @Override
+    public String getFunctionsUri() {
         return this.uri;
     }
 
+    public String getUri() {
```
ahmedmuhsin: This isn't being called, can it be removed?

ahmedmuhsin (commented):
What's the host-side timeline for emitting `https://` URIs? The current host hardcodes `http://127.0.0.1:port/` in [`AspNetCoreGrpcServer.cs`](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs#L35), so this PR's TLS branch is dormant in production today. Is there a tracked host-side change to enable TLS that this PR is gating, or is this defense-in-depth ahead of any host plans? Either is fine, just helps reviewers calibrate.

larohra:
> What's the host-side timeline for emitting `https://` URIs? The current host hardcodes `http://127.0.0.1:port/` in [`AspNetCoreGrpcServer.cs`](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script.Grpc/Server/AspNetCoreGrpcServer.cs#L35), so this PR's TLS branch is dormant in production today. Is there a tracked host-side change to enable TLS that this PR is gating, or is this defense-in-depth ahead of any host plans? Either is fine, just helps reviewers calibrate.

Great questions. I am not aware of the host-level timelines, this is just preparing for that in advance and resolving our own SFIs that come out of it. I'm guessing the Host team has their own SFI items that would get them to resolve this sooner or later.

ahmedmuhsin (approved):

vrdmr (approved):

PR #878, opened 2026-06-16
Generate gRPC test TLS material at runtime instead of committing it

ahmedmuhsin:
## Summary

The gRPC transport tests shipped a self-signed `localhost` certificate, its **private key**, and a **PKCS12 truststore** under `src/test/resources/grpc-tls/`. Committing a PEM private key + PKCS12 bundle trips secret scanning and **hard-blocks the internal ADO mirror push**:

```
VS403654:NonbypassableBlock
  SEC101/013 : PemPrivateKey                     -> localhost-key.pem
  SEC101/055 : Pkcs12CertificatePrivateKeyBundle -> localhost-truststore.p12
```

This replaces the committed key material with **runtime generation**, so no private keys live in the repo.

## What changed

- **New `TestTlsMaterial`** — generates an ephemeral `CN=localhost` keypair, certificate, and matching PKCS12 truststore at test runtime using the JDK's `keytool`. `keytool` ships with every JDK (8–25), so this stays portable across the CI Java matrix without adding a certificate-generation dependency (e.g. Bouncy Castle) and without `--add-exports` for `sun.security` internals. Material is written to a temp dir deleted on JVM exit, generated once per JVM.
- **`FunctionsTestHost`** — builds the server `SslContext` from the generated PKCS12 keystore (`PrivateKey` + `X509Certificate[]`) instead of PEM resource files.
- **`GrpcTransportTest`** — points the client truststore at the generated PKCS12 truststore instead of the committed resource.
- **Deleted** `src/test/resources/grpc-tls/{localhost-cert.pem,localhost-key.pem,localhost-truststore.p12}`.
- **`.gitignore`** guards (`src/test/resources/grpc-tls/`, `*.p12`, `*.pfx`, `*-key.pem`) so keys/keystores can't be re-committed. Verified no currently-tracked files match these patterns.

## Testing

```
mvn test -Dtest=GrpcTransportTest
-> Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```

Covers all three existing cases against the runtime-generated cert:
- legacy plaintext transport
- trusted HTTPS/TLS handshake
- HTTPS does not downgrade to plaintext when TLS fails

## Important caveat

This removes the files from the **branch tip** and prevents recurrence, but it does **not** on its own unblock the mirror. The hard-block is reported at the commit that *introduced* the key (`079a4d40`), which still exists in the history the mirror replays. Fully clearing the mirror still requires **either**:

1. a history rewrite to purge those blobs from every mirrored ref (`dev`, `release/*`, etc.) + force-push, **or**
2. a 1ES secret-scanning exception for those fingerprints/path.

This PR is the clean forward-fix that makes the eventual history purge final and stops new keys from landing.


src/test/java/com/microsoft/azure/functions/worker/functional/tests/GrpcTransportTest.java
```diff
@@ -1,6 +1,7 @@
 package com.microsoft.azure.functions.worker.functional.tests;
 
 import java.net.*;
+import java.nio.file.*;
```
swapnil-nagar: Avoid * imports

Here and in entire PR.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/FunctionsTestHost.java
```diff
@@ -67,16 +67,30 @@ private int populatePort() {
...
-                ? NettyServerBuilder.forPort(this.getPort())
-                    .sslContext(GrpcSslContexts.forServer(getTlsResource(TLS_CERTIFICATE_RESOURCE), getTlsResource(TLS_PRIVATE_KEY_RESOURCE)).build())
+                ? NettyServerBuilder.forPort(this.getPort()).sslContext(buildServerSslContext())
                 : ServerBuilder.forPort(this.getPort());
         this.grpcHost = new HostGrpcImplementation();
         this.server = builder.addService(this.grpcHost).build();
         this.server.start();
     }
 
+    private static SslContext buildServerSslContext() throws Exception {
```
swapnil-nagar: I don't think it needs to be static.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/FunctionsTestHost.java
```diff
@@ -67,16 +67,30 @@ private int populatePort() {
     }
 
     @PostConstruct
-    private void initializeServer() throws IOException {
+    private void initializeServer() throws Exception {
```
swapnil-nagar: This is a broader Exception you are throwing, can you throw specific exception.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/TestTlsMaterial.java
```diff
@@ -0,0 +1,176 @@
...
+        return this.trustStore;
+    }
+
+    private static void generateKeyPair(Path keyStore) throws IOException, InterruptedException {
+        runKeytool(
+                "-genkeypair",
+                "-alias", ALIAS,
+                "-keyalg", "RSA",
+                "-keysize", "2048",
+                "-validity", "7300",
```
swapnil-nagar: 7300 days, that's an eternity. What is the security recommendation?

src/test/java/com/microsoft/azure/functions/worker/test/utilities/TestTlsMaterial.java
```diff
@@ -0,0 +1,176 @@
...
+                "-importcert",
+                "-noprompt",
+                "-alias", ALIAS,
+                "-file", certificate.toString(),
+                "-keystore", trustStore.toString(),
+                "-storetype", STORE_TYPE,
+                "-storepass", PASSWORD);
+    }
+
+    private static void runKeytool(String... arguments) throws IOException, InterruptedException {
```
swapnil-nagar, line 128: runKeytool throws IllegalStateException but method signature does not show it.

Problem:  the constructor only catches IOException | InterruptedException. So timeout or keytool failure will skip the custom wrapper message.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/TestTlsMaterial.java
```diff
@@ -0,0 +1,176 @@
...
+        if (process.exitValue() != 0) {
+            synchronized (output) {
+                throw new IllegalStateException(
+                        "keytool " + arguments[0] + " failed (exit " + process.exitValue() + "):\n" + output);
+            }
+        }
+    }
+
+    private static String keytoolPath() {
+        String javaHome = System.getProperty("java.home");
```
swapnil-nagar, line 173: keytoolPath() assumes java.home/bin/keytool exists something like below will make it robust

Path keytool = Paths.get(javaHome, "bin", windows ? "keytool.exe" : "keytool");
if (!Files.isExecutable(keytool)) {
    throw new IllegalStateException("keytool not found or not executable at: " + keytool);
}

src/test/java/com/microsoft/azure/functions/worker/functional/tests/GrpcTransportTest.java
ahmedmuhsin: Replaced the wildcard imports this PR added with explicit ones and dropped an unused java.net import. The remaining wildcards are pre-existing in these test files.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/FunctionsTestHost.java
ahmedmuhsin: Made it an instance method.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/FunctionsTestHost.java
ahmedmuhsin: Narrowed to GeneralSecurityException and IOException.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/TestTlsMaterial.java
ahmedmuhsin: Reduced to 2 days. It only needs to outlive a single test run.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/TestTlsMaterial.java
ahmedmuhsin: Switched the timeout and non-zero-exit paths to IOException so all failures are wrapped in the same message.

src/test/java/com/microsoft/azure/functions/worker/test/utilities/TestTlsMaterial.java
ahmedmuhsin: Added a Files.isExecutable check before invoking keytool, throwing IOException so it routes through the same wrapper.

swapnil-nagar (approved):

gavin-aguiar (approved):

PR #819, opened 2025-04-22
Add OpenTelemetry support for Java Azure Functions

ahmedmuhsin:
This feature adds a new app settings 
- `JAVA_ENABLE_OPENTELEMETRY`
- `JAVA_APPLICATIONINSIGHTS_ENABLE_TELEMETRY`

When either is set to `true`, the worker informs the host that it supports otel which prevents the duplication of logs by the host.

### Related PR
https://github.com/Azure/azure-functions-java-additions/pull/40

### Work Item 
https://github.com/Azure/azure-functions-pyfx-planning/issues/708

### Pull request checklist

* [ ] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
  * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] My changes **do not** need to be backported to a previous version
  * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [ ] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

PR #880, opened 2026-07-22
Fix functionInstanceInjector null on concurrent cold start (#879)

### LINKED ISSUE #879: Regression: functionInstanceInjector left null on concurrent cold-start load (2.19.2+) — reintroduces the NPE fixed by #684
## Summary

Since worker **2.19.2**, `JavaFunctionBroker.initializeOneTimeLogics()` sets `oneTimeLogicInitialized = true` **before** calling `initializeFunctionInstanceInjector()`. Because function-load requests are dispatched concurrently, a second load thread can observe the flag as already `true`, skip the double-checked-locked init block, and complete while `functionInstanceInjector` is still null. The field then stays null for the life of the worker, and every invocation fails in `ExecutionContextDataSource.getFunctionInstance()`:

```
java.lang.NullPointerException: Cannot invoke
  "com.microsoft.azure.functions.spi.inject.FunctionInstanceInjector.getInstance(java.lang.Class)"
  because "this.functionInstanceInjector" is null
    at com.microsoft.azure.functions.worker.binding.ExecutionContextDataSource.getFunctionInstance(ExecutionContextDataSource.java:108)
    at com.microsoft.azure.functions.worker.broker.JavaMethodInvokeInfo.invoke(JavaMethodInvokeInfo.java:20)
    ...
```

This is the same NPE that was fixed in #684 (which moved the flag assignment to *after* injector init). The 2.19.2 change reordered it back.

## Affected versions

- **Good:** 2.19.1 and earlier (`initializeFunctionInstanceInjector()` runs, *then* `oneTimeLogicInitialized = true`).
- **Regressed:** 2.19.2, 2.19.3, 2.19.4 (flag set before injector init).

Confirmed by reading `JavaFunctionBroker.java` at each tag.

## Root cause (current `main` / 2.19.4)

`initializeOneTimeLogics()`:
```java
if (!oneTimeLogicInitialized) {
    synchronized (oneTimeLogicInitializationLock) {
        if (!oneTimeLogicInitialized) {
            ...
            oneTimeLogicInitialized = true;          // <-- set before init
            initializeFunctionInstanceInjector();    // <-- slow ServiceLoader scan
        }
    }
}
```

`JavaWorkerClient$StreamingMessagePeer.onNext` submits every message (incl. per-function `FunctionLoadRequest`s) to `Executors.newCachedThreadPool()`, so multiple `loadMethod` calls run concurrently at cold start. Thread A enters the block and sets the flag; thread B sees the flag `true` and returns from `initializeOneTimeLogics()` immediately; if B's invocation is dispatched before A finishes the `ServiceLoader.load(FunctionInstanceInjector.class)` scan, the injector is still null. The scan is slower for apps with a large `lib/` classpath and a custom injector provider (e.g. Spring Cloud Function's `spring-cloud-function-adapter-azure` `AzureFunctionInstanceInjector`), widening the window.

## Impact

An affected instance fails ~100% of invocations until recycled; healthy instances coexist, so it looks intermittent in aggregate. Especially damaging for non-HTTP triggers that fire immediately at cold start (e.g. SQL trigger with a pending change), and for always-on / elastic plans where cold starts happen on deploy, restart, and scale/patch events.

## Suggested fix

Restore the #684 ordering: assign `oneTimeLogicInitialized = true` **after** `initializeFunctionInstanceInjector()` completes (and after any other one-time init), so a concurrent thread cannot observe the flag as set while the injector is still null. Consider also initializing the injector eagerly/synchronously before any load request is acked.

## Environment

- Java worker 2.19.4 (host 4.1051.300), Linux, Java 21.
- Spring Cloud Function Azure adapter (custom `FunctionInstanceInjector` provider).

## Workaround

Pin the Functions host to a worker <= 2.19.1 via `linuxFxVersion` (`mcr.microsoft.com/azure-functions/java:4.1040.300-0-java21[-appservice]`).

ahmedmuhsin:
## What this fixes

Fixes #879. Since worker 2.19.2, an instance can start up with `functionInstanceInjector` left null for its whole lifetime, and then every invocation on that instance fails with an NPE.

```
java.lang.NullPointerException: Cannot invoke
  "com.microsoft.azure.functions.spi.inject.FunctionInstanceInjector.getInstance(java.lang.Class)"
  because "this.functionInstanceInjector" is null
    at com.microsoft.azure.functions.worker.binding.ExecutionContextDataSource.getFunctionInstance(ExecutionContextDataSource.java:108)
```

## Root cause

`JavaFunctionBroker.initializeOneTimeLogics()` uses double-checked locking behind the volatile `oneTimeLogicInitialized` flag. Since 2.19.2 it set the flag to true before building the injector.

```java
oneTimeLogicInitialized = true;          // flag set first
initializeFunctionInstanceInjector();    // slow ServiceLoader scan runs after
```

Function load requests are dispatched concurrently, since `JavaWorkerClient` submits every message to a cached thread pool. At cold start two threads run `loadMethod` at the same time. Thread A enters the locked block and sets the flag. Thread B hits the lock-free outer check, sees the flag already true, returns early, and starts invoking while the injector is still null. The `ServiceLoader` scan for a custom injector (for example the Spring Cloud Function adapter) widens the window.

This is the same NPE that #684 fixed. #819 (OpenTelemetry support) reordered the two lines and reopened it.

## The fix

Build the injector before flipping the flag, which restores the #684 ordering.

```java
initializeFunctionInstanceInjector();
oneTimeLogicInitialized = true;
```

## Test

Adds `JavaFunctionBrokerConcurrentInitTest`, which freezes a thread at the point the injector is about to be built and asserts `oneTimeLogicInitialized` is still false there. It fails on the old ordering and passes with the fix.

## Affected versions

Regressed in 2.19.2 and present through 2.19.4. 2.19.1 and earlier are fine.

### DIFF
--- a/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
+++ b/src/main/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBroker.java
@@ -115,8 +115,12 @@ public class JavaFunctionBroker {
 						initializeInvocationChainFactory();
 					}
 
-					oneTimeLogicInitialized = true;
+					// Initialize the injector before flipping the flag. The outer check on
+					// oneTimeLogicInitialized is lock-free, so a concurrent load thread that sees the
+					// flag set will skip init and can invoke while functionInstanceInjector is still
+					// null. See issue #879.
 					initializeFunctionInstanceInjector();
+					oneTimeLogicInitialized = true;
 				}
 			}
 		}
--- a/src/test/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBrokerConcurrentInitTest.java
+++ b/src/test/java/com/microsoft/azure/functions/worker/broker/JavaFunctionBrokerConcurrentInitTest.java
@@ -1,0 +1,83 @@
+package com.microsoft.azure.functions.worker.broker;
+
+import com.microsoft.azure.functions.worker.reflect.ClassLoaderProvider;
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Regression guard for issue #879.
+ *
+ * initializeOneTimeLogics() runs the double-checked locking pattern behind a volatile flag. The
+ * injector must be built before that flag flips to true. If the flag is set first, a second load
+ * thread can pass the lock-free outer check, skip initialization, and start invoking while
+ * functionInstanceInjector is still null, which throws NPE in
+ * ExecutionContextDataSource.getFunctionInstance().
+ */
+public class JavaFunctionBrokerConcurrentInitTest {
+
+    @Test
+    public void flagIsNotSetUntilInjectorIsInitialized() throws Exception {
+        CountDownLatch reachedInjectorInit = new CountDownLatch(1);
+        CountDownLatch releaseInjectorInit = new CountDownLatch(1);
+        AtomicInteger createClassLoaderCalls = new AtomicInteger();
+
+        // createClassLoader() is called once at the top of initializeOneTimeLogics() and once inside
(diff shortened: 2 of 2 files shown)


TsuyoshiUshio (approved):
LGTM. 

One tip for `CountDownLatch` If it works, that's great, but when I use similar solution in C#, CI has very few thread, so that blocked long and eventually timed out. But It is also possible not the case. Let's merge it and see if it works. If it works, I don't have any objection and CountDownLatch will remove flaky testing.

swapnil-nagar (approved):

PR #893, opened 2026-08-13
[experiment] ci: try 1es-ubuntu-24.04-min for Linux jobs instead of disk cleanup

ahmedmuhsin:
**Experiment — not for merge as-is.** Opened as a draft to get a CI signal.

## What
Switch the two Linux CI jobs (emulated tests, docker integration tests) from `1es-ubuntu-22.04` to `1es-ubuntu-24.04-min`, and drop the `Free up disk space` cleanup steps that #883 added. A `df`/`du` report step is kept so the run tells us the new image's baseline headroom.

## Why
The `1es-ubuntu-22.04` image boots ~92% full on `/` (67G used of 73G), almost entirely from toolsets these jobs never touch — Android SDK and Haskell under `/usr/local` (~24G), pre-cached docker images (~12G), Swift/miniconda under `/usr/share` (~9.5G), CodeQL/PyPy in the tool cache (~6.7G). That left so little headroom that `dotnet test` filled the disk, the agent process aborted with exit 134, and ADO reported the job as **canceled**.

#883 worked around it by `rm -rf`-ing those toolsets at job start (`/` 92% → 54%, ~28G freed), which has held: 17 of 18 nightly `dev` builds since have passed with zero cancellations.

That workaround is still deleting things out from under a hosted image, which is fragile — it silently breaks if paths move in a future image. If the `-min` image ships lean enough, we can drop the cleanup entirely and rely on the image instead.

## What to look for
1. Do the jobs still provision? `-min` images carry fewer preinstalled toolsets, so the likely failure mode is a missing dependency (Maven, preinstalled JDKs for `JavaToolInstaller`, docker, Python) rather than disk.
2. The `Report disk space` step output — how much free space does `-min` actually start with?

If it provisions cleanly with comfortable headroom, the follow-up is to drop the cleanup steps for good. If it needs extra tool installs, we compare that cost against keeping the `rm -rf`.

## Scope
CI only, Linux only. Untouched: the Windows jobs, `java-version-check.yml`, and the official `run-e2e-tests-linux.yml` — those stay on `1es-ubuntu-22.04` until this experiment reports back.

### DIFF
--- a/eng/ci/templates/jobs/run-docker-tests-linux.yml
+++ b/eng/ci/templates/jobs/run-docker-tests-linux.yml
@@ -9,7 +9,7 @@ jobs:
 
     pool:
       name: ${{ parameters.poolName }}
-      image: 1es-ubuntu-22.04
+      image: 1es-ubuntu-24.04-min
       os: linux
 
     strategy:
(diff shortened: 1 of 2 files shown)


ahmedmuhsin:
Superseded by #895, which now carries this change together with the Docker Hub packaging fix. The emulated Linux job on 1es-ubuntu-24.04-min is unchanged from the version validated here (all 5 legs green, 62G free with no cleanup step); the docker legs that were red on this PR were the pre-existing dev breakage that #895 fixes.

